PR #1, opened 2016-07-07
Apply patch from POC repository

alexmiranda:
Signed-off-by: Alex Miranda <alexmirandamoraes@gmail.com>


src/createComponent.js
```diff
@@ -22,16 +22,12 @@ export default function createComponent(options = {}) {
       if (typeof options.afterMount === 'function') {
         return options.afterMount.call(this);
       }
-
-      return null;
```
fahad19: please keep them, ESLint would fail otherwise.

alexmiranda: excluded createComponent.js from patch and reapplied it

PR #2, opened 2016-07-08
Fix rendering twice issue

alexmiranda:
Issue:
Region component was rendered twice because of `setState` being called during `componentDidMount`

Fix
Change `componentDidMount` to `componentWillMount`.

### DIFF
--- a/src/components/Region.js
+++ b/src/components/Region.js
@@ -19,7 +19,7 @@ export default React.createClass({
     name: PropTypes.string.isRequired
   },
 
-  componentDidMount() {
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "frint",
-  "version": "0.2.4",
+  "version": "0.2.5",
```
fahad19: please don't update the version numbers in PRs. we do it via `npm version` while releasing separately :)

alexmiranda: removed

PR #5, opened 2016-07-12
Adds the createFactory spec

mAiNiNfEcTiOn:
# What does this PR do:
  - Adds the usage of `sinon.sandbox` to create and restore stubs/spies, to not pollute other tests.
  - Adds a new test for the `createFactory()`, only testing the result expectations.

screen shot 2016-07-12 at 10 47 02

  - Diffs
  - Run `npm test`

  - Adds the test `test/createFactory.spec.js`.

### DIFF
--- a/test/createFactory.spec.js
+++ b/test/createFactory.spec.js
@@ -10,7 +10,7 @@ describe('createFactory', function () {
     const fakeAppInstance = {};
     const mySpec = {
       initialize: sinon.stub(),
-      myCustomFunction: () => 'lol'
+      myCustomFunction: () => 'value'
     };
 
     const MyFactory = createFactory(mySpec);
@@ -25,6 +25,6 @@ describe('createFactory', function () {
 
     /* And contain the functions passed in the spec */
     expect('myCustomFunction' in myFactoryInstance).to.be.equal(true);
-    expect(myFactoryInstance.myCustomFunction()).to.be.equal('lol');
+    expect(myFactoryInstance.myCustomFunction()).to.be.equal('value');
   });
 });
--- a/test/createFactory.spec.js
+++ b/test/createFactory.spec.js
@@ -6,24 +6,25 @@ import createFactory from '../src/createFactory';
 chai.use(sinonChai);
 
 describe('createFactory', function () {
-  it('returns a Factory class', function () {
-    const fakeAppInstance = {};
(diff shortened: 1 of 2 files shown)


test/createFactory.spec.js
```diff
@@ -0,0 +1,30 @@
+/* global describe, it */
+import chai, { expect } from 'chai';
+import sinon from 'sinon';
+import sinonChai from 'sinon-chai';
+import createFactory from '../src/createFactory';
+chai.use(sinonChai);
+
+describe('createFactory', function () {
+  it('returns a Factory class', function () {
```
alexmiranda: if you instead use a nested `describe` function here, or alternatively, move the initialisation part of the test i.e lines 10-17, to either the parent `describe('createFactory', ...)` or a `before` method, you can make this test more readable by extracting each of the assertion groups, i.e. lines 19-20; 22-24; and 26-28 into their own methods.  The reporting will be a lot better too, because it will be more clear what the expectations are and a failing test would tell you straight away what's failed, without you needing to go over a long list of assertions to see what's wrong.

it should be similar to this (did not test this code though):

```js
describe('createFactory', () => {
  const fakeAppInstance = {};
  const mySpec = {
    initialize: sinon.stub(),
    myCustomFunction: () => 'value'
  };

  const MyFactory = createFactory(mySpec);
  const myFactoryInstance = new MyFactory({ app: fakeAppInstance });

  it('should execute the intialize() at construction', () => {
    expect(mySpec.initialize).to.be.callCount(1);
  });

  it('should be an instance of MyFactory and contain the app instance passed', () => {
    expect(myFactoryInstance).to.be.instanceOf(MyFactory);
    expect(myFactoryInstance.app).to.be.deep.equal(fakeAppInstance);
  });

  it('should contain the functions passed in the spec', () => {
    expect('myCustomFunction' in myFactoryInstance).to.be.equal(true);
    expect(myFactoryInstance.myCustomFunction()).to.be.equal('value');
  });
});
```

mAiNiNfEcTiOn: Indeed ... will do it. Thanks @alexmiranda

mAiNiNfEcTiOn: Done 👍 nice input.

alexmiranda:
Nice work. LGTM!

PR #6, opened 2016-07-12
Improves the createComponent.spec.js

mAiNiNfEcTiOn:
# What does this PR do:
  - Changes the `test/createComponent.spec.js` to make the code's behavior more _readable_ (thanks @alexmiranda).

screen shot 2016-07-12 at 14 18 18

  - Diffs

Updated the `test/createComponent.spec.js`.

### DIFF
--- a/test/createComponent.spec.js
+++ b/test/createComponent.spec.js
@@ -6,27 +6,27 @@ import sinonChai from 'sinon-chai';
 import createComponent from '../src/createComponent';
 
 const sandbox = sinon.sandbox.create();
-
 chai.use(sinonChai);
 
-
 describe('createComponent', function () {
+  const mySpec = {
+    myCustomFunction() { return 'foo'; },
+    render() { return null; }
+  };
+  let MyComponent;
+  let myComponentInstance;
+
   beforeEach(function () {
     sandbox.spy(React, 'createClass');
+    MyComponent = createComponent(mySpec);
+    myComponentInstance = new MyComponent();
   });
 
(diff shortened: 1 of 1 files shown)


test/createComponent.spec.js
```diff
@@ -26,8 +36,14 @@ describe('createComponent', function () {
...
       });
-    expect(myComponentInstance).to.be.instanceof(MyComponent);
+  });
+
+  it('is a valid React component and a MyComponent\'s instance', function () {
     expect('isReactComponent' in Object.getPrototypeOf(myComponentInstance)).to.be.equal(true);
+    expect(myComponentInstance).to.be.instanceof(MyComponent);
+  });
+
+  it('has the spec\'s functions', function () {
```
markvincze: This is not an actual review comment, just a Javascript question: isn't 
```
"has the spec's functions"
```
nicer? Or the linting doesn't allow double quotes at all?

markvincze:
:+1:

test/createComponent.spec.js
mAiNiNfEcTiOn: Linting rule: Strings should be either single quoted or, in case we need to concatenate with variables, we should use template literals.

But I can improve it to be 'has the functions of the spec'

alexmiranda:
LGTM

test/createComponent.spec.js
markvincze: Oh no, It's fine by me, I was just curious

PR #8, opened 2016-07-13
Adds the createModel.spec.js

mAiNiNfEcTiOn:
# What does this PR do:
  - Adds the createModel.spec.js.

screen shot 2016-07-13 at 09 28 53

  - Diffs
  - `npm test`

Added the test `createModel.spec.js`

### DIFF
--- a/test/createModel.spec.js
+++ b/test/createModel.spec.js
@@ -1,0 +1,45 @@
+/* global describe, it */
+import chai, { expect } from 'chai';
+import sinon from 'sinon';
+import sinonChai from 'sinon-chai';
+
+import BaseModel from '../src/Model';
+import createModel from '../src/createModel';
+chai.use(sinonChai);
+
+describe('createModel', () => {
+  const mySpec = {
+    getAttribute1() { return this.attributes.attribute1; },
+    getAttribute2() { return this.attributes.attribute2; }
+  };
+
+  const myAttributes = {
+    attribute1: 'value1',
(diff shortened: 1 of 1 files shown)


test/createModel.spec.js
```diff
@@ -0,0 +1,45 @@
+/* global describe, it */
+import chai, { expect } from 'chai';
+import sinon from 'sinon';
+import sinonChai from 'sinon-chai';
```
fahad19: we need to load sinon in this spec?

test/createModel.spec.js
```diff
@@ -0,0 +1,45 @@
...
+
+import BaseModel from '../src/Model';
+import createModel from '../src/createModel';
+chai.use(sinonChai);
```
alexmiranda: This is becoming repetitive here. Can we use `--require` on `mocha.opts` to load a helper or something? It doesn't need to be in this PR, just wanted to know your opinion and I can make the change in a separate PR along with other tests.

test/createModel.spec.js
mAiNiNfEcTiOn: Removing sir.

test/createModel.spec.js
mAiNiNfEcTiOn: Agree, but then again, you don't need to load it all the time :) This is a case ;-)

alexmiranda:
LGTM

PR #13, opened 2016-07-15
Adds createService spec

mAiNiNfEcTiOn:
# What does this PR do:
  - Converts `function ()` to `() =>` in all specs that were using them.
  - Adds a test in the `createFactory.spec.js` to test if the constructor throws an error when no app instance is provided.
  - Adds the `createService.spec.js`.

screen shot 2016-07-15 at 18 34 01

  - Diffs.
  - `npm test`

All tests that were using anonymous functions were converted to arrow functions.
Added the test `test/createService.spec.js`

### DIFF
--- a/test/createFactory.spec.js
+++ b/test/createFactory.spec.js
@@ -5,7 +5,7 @@ import sinonChai from 'sinon-chai';
 import createFactory from '../src/createFactory';
 chai.use(sinonChai);
 
-describe('createFactory', function () {
+describe('createFactory', () => {
   const fakeAppInstance = {};
   const mySpec = {
     initialize: sinon.stub(),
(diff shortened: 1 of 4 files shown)


test/createService.spec.js
```diff
@@ -0,0 +1,48 @@
...
+  it('creates instance from my service', () => {
+    expect(myServiceInstance).to.be.instanceOf(MyService);
+  });
+
+  it('executes the initialize() at construction', () => {
+    expect(mySpec.initialize).to.be.callCount(1);
+  });
+
+  it('must contain the functions passed in the spec', () => {
+    expect('initialize' in myServiceInstance).to.be.equal(true);
```
alexmiranda: `expect(myServiceInstance).to.include.keys('initialize');`

test/createService.spec.js
```diff
@@ -0,0 +1,48 @@
...
+  it('throws error when an app instance is not provided', () => {
+    expect(() => new MyService()).to.throw(Error, 'App instance not provided.');
+  });
+
+  it('creates instance from my service', () => {
+    expect(myServiceInstance).to.be.instanceOf(MyService);
+  });
+
+  it('executes the initialize() at construction', () => {
+    expect(mySpec.initialize).to.be.callCount(1);
```
alexmiranda: using `chai-spies`:
`expect(mySpec.initialize).to.have.been.called.once`
what do you think?

mAiNiNfEcTiOn: We had that before on our internal repo, but since we're in favour of functions instead of expressions ( you can see it based on our eslint rule `no-unused-expressions` ).

Also, we had some problems where the `called.once` was not validating properly ( sometimes the callCount was bigger than 1 ).

I would prefer to stay away from expressions and use functions since we're pretty much into functional programming/pure functions as well :)

test/createService.spec.js
mAiNiNfEcTiOn: Yup yup. Haven't been very familiar with Chai's APIs, this one seems really nice :)

Changing.

test/createFactory.spec.js
```diff
@@ -15,16 +15,20 @@ describe('createFactory', function () {
   const MyFactory = createFactory(mySpec);
   const myFactoryInstance = new MyFactory({ app: fakeAppInstance });
 
-  it('executes the initialize() at construction', function () {
+  it('throws error when an app instance is not provided', () => {
+    expect(() => new MyFactory()).to.throw(Error, 'App instance not provided.');
```
markvincze: :+1:

markvincze:
:+1:

test/createService.spec.js
alexmiranda: https://www.npmjs.com/package/dirty-chai

mAiNiNfEcTiOn: @alexmiranda now that's what I'm talking about :D

PR #15, opened 2016-07-15
Adds the render spec

mAiNiNfEcTiOn:
# What does this PR do:
  - Changes the `test/setup.js` to add another DOM element ( `<div id="root"></div>` to be used for rendering components (avoiding to render in the `body` element and getting a warning).
  - Adds the `test/render.spec.js`

screen shot 2016-07-15 at 21 59 56

  - Diffs
  - `npm test`

  - Adds spec `test/render.spec.js`

### DIFF
--- a/test/render.spec.js
+++ b/test/render.spec.js
@@ -18,7 +18,7 @@ describe('render', () => {
     afterMount: sandbox.stub(),
     beforeMount: sandbox.stub(),
     beforeUnmount: sandbox.stub(),
-    render: sandbox.stub().returns(fakeComponent)
+    render: sandbox.stub().returns(FakeComponent)
   };
 
   let renderedComponent;
--- a/test/render.spec.js
+++ b/test/render.spec.js
@@ -7,7 +7,7 @@ import sinonChai from 'sinon-chai';
 chai.use(sinonChai);
 
 
-const fakeComponent = React.createClass({
+const FakeComponent = React.createClass({
   render() { return null; }
 });
 const targetElement = document.getElementById('root');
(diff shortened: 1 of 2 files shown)


test/render.spec.js
```diff
@@ -0,0 +1,64 @@
+import chai, { expect } from 'chai';
+import render from '../src/render';
+import React from 'react';
+import ReactDOM from 'react-dom';
+import sinon from 'sinon';
+import sinonChai from 'sinon-chai';
+chai.use(sinonChai);
+
+
+const fakeComponent = React.createClass({
```
alexmiranda: maybe use initial upper case here? i.e.:
`const FakeComponent = React.createClass({`

test/setup.js
```diff
@@ -1,6 +1,6 @@
 import { jsdom } from 'jsdom';
 
-global.document = jsdom('<html><body></body></html>');
+global.document = jsdom('<html><body><div id="root"></div></body></html>');
```
alexmiranda: 👍

alexmiranda:
Nice work! I would just consider improving the readability of some tests by using `to.have.been.called()`, `to.have.been.calledOnce()` and `to.have.been.calledWith(args)` .  `callCount` can lead to the spread of magic numbers in these and other tests in future, so I'd avoid it.

test/render.spec.js
mAiNiNfEcTiOn: Indeed :) fixing it 👍

mAiNiNfEcTiOn:
@alexmiranda good point.

Yes, indeed it improves readability. Will try to improve them - in all specs - in another PR :)
As long as it is function-based call, it's super fine

markvincze:
:+1:

alexmiranda:
LGTM. :+1:

PR #17, opened 2016-07-18
Expose PropTypes

fahad19:
## What's done:

Now you can import and use PropTypes as follows:

```js
import { createComponent, PropTypes } from 'frint';

const MyComponent = createComponent({
  propTypes: {
    name: PropTypes.string.isRequired
  },

  render() {
    return (
      <p>Name is {this.props.name}</p>
    );
  }
});
```

## Unit and/or functional tests:

Yes

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -27,6 +27,7 @@
 * [API Reference](/docs/api/README.md)
   * [createApp](/docs/api/createApp.md)
   * [createComponent](/docs/api/createComponent.md)
+  * [PropTypes](/docs/api/PropTypes.md)
   * [combineReducers](/docs/api/combineReducers.md)
   * [createModel](/docs/api/createModel.md)
   * [createService](/docs/api/createService.md)
--- a/docs/api/Component.md
+++ b/docs/api/Component.md
@@ -1,3 +1,19 @@
 # Component
 
 Component classes ready to be embedded as JSX for rendering.
+
+## Usage
+
+```js
+import { createComponent } from 'frint';
+
+const MyComponent = createComponent({
+  render() {
+    return (
+      <p>Name is {this.prop.name}</p>
+    );
+  }
+});
+```
+
+It is best practice to define `propTypes` while defining the Component too. You can read more about it in [PropTypes](./PropTypes.md) section.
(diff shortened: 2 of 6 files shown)


alexmiranda:
LGTM. :+1:

mAiNiNfEcTiOn:
Hmmm.. That doesn't sound cool. So you're basically bringing the concept of _PropTypes_ to what _frint_ exposes.

Unless you're thinking on applying an abstraction layer on top of the _PropTypes_, it makes it very dependent on that API type.
Ideally and since we're on a "pure functions" kind of thing, a similar API would be (when declaring the component):

```javascript
import { createComponent, Types } from 'frint';
export default createComponent({
  /* ... */,
  properties: {
    prop1: Type.isString().isRequired(),
    prop2: Type.isNumber().isRequired(),
    prop3: Type.isBoolean()
  },
  /* ... */
});
```

(This is an example, do not take it literally)

mAiNiNfEcTiOn:
Am I being very _annoying_? I am not fighting over naming, is more about following the _abstraction_ pattern :/

markvincze:
I've taken a look, but I don't have enough knowledge to give meaningful input.

fahad19:
RIght now I am not even thinking of it as an abstraction layer. Just passing it out as `combineReducers` is exported. Otherwise we need to ask other developers to also install `react`, and then get `React.PropTypes`.

I understand your point of abstracting it away, but if we do, I would suggest we wait for the next major (breaking) version of `frint`. A lot of things and our needs might evolve by then.

mAiNiNfEcTiOn:
Well so be it... And if it's a blocking feature for AIG's tasks, do it.

But bear in mind ( 🐻 ) that we're going (with PropTypes) via an _expressions_ route on it... Changing it later to _chainable functions_, will require more.

But, fine by me. Don't want to block anyone.

fahad19:
> But bear in mind ( 🐻 ) that we're going (with PropTypes) via an expressions route on it... Changing it later to chainable functions, will require more.

I tacked the chainable expressions part in a private project of mine this weekend. We can go really deep with chaining via `Object.defineProperty`, and they can still be written as just expressions, without any `()`. Will show you when you come back.

PR #23, opened 2016-08-16
Adds getValidationFunctions() method to createApp.js

mAiNiNfEcTiOn:
# What does this PR do:

Adds the method `getValidationFunctions()` to `createApp.js`

Removes the method `getDOMElement()` from `createComponent.js`

screen shot 2016-08-16 at 12 40 20

  - Diffs
  - `npm run test`

- 3 tests added to `createApp.spec.js`
- 1 test removed from `createComponent.spec.js`

### DIFF
--- a/src/createApp.js
+++ b/src/createApp.js
@@ -291,6 +291,23 @@ export default function createApp(options = {}) {
 
       return null;
     }
+
+    /**
+     * Attempts to get the option 'validationFunctions' from the RootApp.
+     * If it's not possible, returns from the current app as a fallback
+     *
+     * @method getValidationFunctions
+     * @return {Array|undefined} List of functions
+     * @public
+     */
+    getValidationFunctions() {
+      const rootApp = this.getRootApp();
+      if (rootApp && rootApp.getOption('validationFunctions')) {
+        return rootApp.getOption('validationFunctions');
+      }
+
+      return this.getOption('validationFunctions');
+    }
   }
 
   return App;
--- a/src/createComponent.js
+++ b/src/createComponent.js
@@ -35,16 +35,5 @@ export default function createComponent(options = {}) {
 
       return null;
     },
-
-    /**
-     * Returns the root HTML element of the component.
-     *
-     * @method getDOMElement
-     * @return {HTMLElement|null} Returns the component's root HTML Node.
-     * @public
-     */
-    getDOMElement() {
-      return ReactDOM.findDOMNode(this);
-    },
   });
 }
--- a/test/createApp.spec.js
+++ b/test/createApp.spec.js
@@ -4,6 +4,14 @@ import { expect } from 'chai';
 import createApp from '../src/createApp';
 
 describe('createApp', () => {
+  beforeEach(() => {
+    window.app = null;
+  });
+
+  afterEach(() => {
+    delete window.app;
+  });
+
   it('creates an instance', () => {
     const App = createApp({
       name: 'MyAppName',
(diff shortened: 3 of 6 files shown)


markvincze:
:+1:

src/createApp.js
```diff
@@ -291,6 +291,23 @@ export default function createApp(options = {}) {
...
+     * Attempts to get the option 'validationFunctions' from the RootApp.
+     * If it's not possible, returns from the current app as a fallback
+     *
+     * @method getValidationFunctions
+     * @return {Array|undefined} List of functions
+     * @public
+     */
+    getValidationFunctions() {
+      const rootApp = this.getRootApp();
+      if (rootApp && rootApp.getOption('validationFunctions')) {
```
alexmiranda: what about avoiding to call `rootApp.getOption` twice?

```js
return rootApp
  ? rootApp.getOption('validationFunctions') || this.getOption('validationFunctions')
  : this.getOption('validationFunctions');
 ```

File: test/createApp.spec.js [hunk 2]
```diff
@@ -16,4 +24,78 @@ describe('createApp', () => {
...
+    const App = createApp({
+      appId: '123',
+      component: true,
+      name: 'MyAppName',
+      validationFunctions: expectedValidationFunctions,
+    });
+
+    const app = new App();
+
+    expect(app).to.be.instanceof(App);
```
alexmiranda: this expectation is not really what this test verifies, we can perfectly live without it.

test/createApp.spec.js
```diff
@@ -16,4 +24,78 @@ describe('createApp', () => {
...
+    const ChildApp = createApp({
+      appId: '234',
+      component: true,
+      name: 'MyChildAppName',
+      validationFunctions: unExpectedValidationFunctions
+    });
+    const childApp = new ChildApp();
+
+
+    expect(rootApp).to.be.instanceof(RootApp);
```
alexmiranda: I'd say that only the last two assertions are important here, the other _sanity_ checks, i.e. `instanceof` and `eql`, are doing nothing, but testing the test code itself.  Make it shorter and the readability will be better.

test/createApp.spec.js
```diff
@@ -16,4 +24,78 @@ describe('createApp', () => {
...
+
+    const ChildApp = createApp({
+      appId: '234',
+      component: true,
+      name: 'MyChildAppName',
+    });
+    const childApp = new ChildApp();
+
+
+    expect(rootApp).to.be.instanceof(RootApp);
```
alexmiranda: same as above.

test/createComponent.spec.js
```diff
@@ -47,10 +46,6 @@ describe('createComponent', () => {
     expect(myComponentInstance).to.be.instanceof(MyComponent);
   });
 
-  it('gets the DOM Node when executing getDOMElement()', () => {
```
alexmiranda: why remove this test?

src/createComponent.js
```diff
@@ -35,16 +35,5 @@ export default function createComponent(options = {}) {
 
       return null;
     },
-
```
alexmiranda: why remove this method?

mAiNiNfEcTiOn:
Thanks for the comments but we are moving to a service.

I will then leave the method in frint but requires further investigation on why we are getting this type of error: http://blog.j0.hn/post/136543754481/react-cant-find-the-root-component-node-a

(ignore the server-side rendering, focus on the error though :D)

PR #24, opened 2016-08-17
Add code coverage with Coveralls.io

mAiNiNfEcTiOn:
# What does this PR do:
Adds integration with Coveralls.io via TravisCI.
Adds 2 new dependencies: `istanbul` and `coveralls` which allow us to get code coverage stats and send them to coveralls.io

screen shot 2016-08-17 at 11 20 43
screen shot 2016-08-17 at 11 20 46
screen shot 2016-08-17 at 13 14 22

  -  Diffs

N/A


jvantillo:
Nice :) Does coveralls enforce a minimum coverage (since it's in a red state?)

mAiNiNfEcTiOn:
@jvantillo doesn't *enforce* it, just warns that some checks were not successful.

You can still see a "merge pull request" button in the button, just it's not green, which means that not all is ok :)

mAiNiNfEcTiOn:
And this percentage is configurable on coveralls.io. For now, I set it to 50%, for obvious reasons. And a 1% threshold when decreasing :P

PR #26, opened 2016-08-18
Do not load logger in production mode

fahad19:
Checks `process.env.NODE_ENV`, and if the value is `production`, it won't load logger.

---

Next: our webpack configs in private repos need to pass `process.env.NODE_ENV` via DefinePlugin. coming soon.


markvincze:
Which logger is this exactly?  
Is this the one producing the state change messages in the Chrome dev console?

fahad19:
Yep, `redux-logger` to be precise.

markvincze:
:+1:

PR #27, opened 2016-08-19
Implement `alex` for linting docs

fahad19:
While reviewing other PRs, I noticed some pages contain some words that we are not comfortable with. Reviewing each word can be painful for everyone, hence this has now been automated.

The package being used: http://alexjs.com/

Whenever a Travis CI build is triggered, it will run a check against all markdown files, and see if we use any sensitive words or not. If we do, it will fail.

To manually test it:

```
$ npm run alex
```


alexmiranda:
LGTM 👍

PR #28, opened 2016-08-19
Make root index.js es5 compatible

fahad19:
We need this so that Babel can be ignored while bundling third-party vendors.

Would make things faster for us.


markvincze:
:+1:

mAiNiNfEcTiOn:
Well... `const` is supported at the current LTS. But not the shorthand for the properties. :)

alexmiranda:
LGTM 👍

PR #31, opened 2016-09-01
Fixes Region to differentiate widgets with the same name but different appNames

mAiNiNfEcTiOn:
# What does this PR do:
When two different apps with the same widget name tried to render in the same Region, only one was rendered.

E.g.

myFirstApp - helloWorldWidget
mySecondApp - helloWorldWidget

This PR fixes that.

(Proof)
screen shot 2016-09-01 at 23 09 09

  - Diffs
  - With the current master, try to create 2 widgets with the same name and different appNames and target the same region. Only 1 will render. The other one will be assumed as rendered (due to [this previous check](https://github.com/Travix-International/frint/compare/fix-region-to-differentiate-apps-with-same-name?expand=1#diff-fbd171f24f3d6b7eb868c635ff42166eL72)).

Tests were added for `components/Region.js`.

### DIFF
--- a/src/components/Region.js
+++ b/src/components/Region.js
@@ -56,6 +56,7 @@ export default React.createClass({
               return;
             }
 
+            const appName = widget.getOption('appName');
             const widgetName = widget.getOption('name');
 
             // @TODO: later re-implement this check with observables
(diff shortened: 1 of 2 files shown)


fahad19:
very nice find! 👍

markvincze:
:+1:

alexmiranda:
LGTM 👍

PR #32, opened 2016-09-05
Fix mixed ES syntax

alexmiranda:
## What's changed

* Moved `index.js` into `src` so it's transpiled as the other files
* Changed main file to `./lib/index.js`
* Converted `index.js` to ES6 
* Added unit tests for `index.js`

Thanks to @paulcodiny for reporting the issue.


src/index.js
```diff
@@ -0,0 +1,26 @@
...
+  createApp,
+  createComponent,
+  createFactory,
+  createModel,
+  createService,
+  mapToProps,
+  Model,
+  PropTypes,
+  Region,
+  render,
```
AlexDudar: not necessary comma

alexmiranda: our lint rules require a trailing comma for ES6 code

alexmiranda:
after adding unit tests, the test coverage _decreased_! 😐

src/index.js
AlexDudar: strange.. can't say that I like it, but fine

alexmiranda: dangle commas help a lot when visualising changes and make it more convenient and less error-prone for manual merges.

fahad19:
👍

PR #33, opened 2016-10-19
Fix: for passing initial shared state across Widgets

fahad19:
## The bug

To reproduce: 

* Create Widget `A` and `B`
* `B` wants to read state of `A`
* When `B`'s container component is rendered, it doesn't have the state from `A` **initially**
* Only after a new action has been fired in `A`'s store, `B` gets the state right way

## The fix

* Instead of only updating the state from inside the subscription, we are now passing the state initially too, to make sure it is available to Container components at all times.

---

## Additional changes

* A new `resetDOM` function in tests to clear globals in mocha
* Unit tests for `mapToProps` function
* Coverage increased from 64% to **83%**

### DIFF
--- a/src/components/mapToProps.js
+++ b/src/components/mapToProps.js
@@ -24,6 +24,10 @@ export default function mapToProps(opts = {}) {
 
   return (Component) => {
     const WrappedComponent = React.createClass({
+      displayName: (typeof Component.displayName !== 'undefined')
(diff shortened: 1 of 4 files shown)


mAiNiNfEcTiOn:
@fahad19 planning to do the unit test for this one?

markvincze:
Unit tests +1 😃

PR #36, opened 2016-10-29
Tests: for `mapToProps`, `createApp`, and middleware

fahad19:
## What's done

* Added tests for mapping Model to props via `mapToProps`
* Added tests for `createApp`
* Added tests for internal `appendAction` middleware
* Add badge to Coveralls in README


fahad19:
Finally hit the 90% mark, and on the greener side of the badge \o/

PR #39, opened 2016-11-01
mapToProps: support new `observe` key for Observables

fahad19:
## Background

* It is difficult for Widgets to communicate with each other
* A Pub Sub model can help solve the problem
* But our `mapToProps` handles re-renders based on state only

Attempted solution:

* If `mapToProps` can trigger re-renders based on Observables, it would help

## What's done

* A new `observe` key has been introduced in `mapToProps`. And it is expected to return an Observable.
* `rxjs` has been upgraded to v5.0.0-rc.1

## Usage example

```js
// ./components/Root.js
import { Observable } from 'rxjs';
import { createComponent, mapToProps } from 'frint';

const RootComponent = createComponent({
  render() {
    return (
      <div>
        <div>{this.props.appId}</div>
        <div>{this.props.total}</div>
      </div>
    );
  }
});

export default mapToProps({
  observe: (app) => {
    // you have access to `app` instance

    return Observable
      // iterate over these values in a sequence
      .of(1, 2)

      // now convert it to an object, which will be available as props
      .scan(
        // this function would be called twice,
        // first with `1`, and then `2` as number
        (props, number) => {
          props.total = props.total + number;

          return props;
        },

        // the initial object to start the `scan` with
        {
          appId: app.getOption('appId'),
          total: 0
        }
      );
  }
})(RootComponent);
```

The example is a very simple one, but given the `observe` function is passed with the `app` instance, you can get observables from other Services/Factories, and combine them into one, and then produce the final `props` object to be made available in the Component.

And every time the observable emits a new value, it would trigger a re-render in your Component.

## Further reading

We already use Observables in `Region`. 

You can read further about it here: http://reactivex.io/rxjs/

### DIFF
--- a/src/components/mapToProps.js
+++ b/src/components/mapToProps.js
@@ -2,6 +2,8 @@ import _ from 'lodash';
 import React, { PropTypes } from 'react';
 import { connect } from 'react-redux';
 
+import isObservable from '../utils/isObservable';
+
 export default function mapToProps(opts = {}) {
   const options = {
     app: () => {},
@@ -19,6 +21,7 @@ export default function mapToProps(opts = {}) {
     services: {},
     shared: () => {},
     state: () => {},
+    observe: null,
     ...opts,
   };
 
@@ -40,7 +43,9 @@ export default function mapToProps(opts = {}) {
 
       componentWillMount() {
         this.storeSubscriptions = {};
+        this.observeSubscription = null;
 
+        // shared state
         this.context.app.readableApps.forEach((readableAppName) => {
           const readableAppStore = this.context.app.getStore(readableAppName);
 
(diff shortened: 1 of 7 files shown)


mAiNiNfEcTiOn (approved):

src/components/mapToProps.js
```diff
@@ -68,6 +73,20 @@ export default function mapToProps(opts = {}) {
           });
         });
 
+        // observe
+        if (typeof options.observe === 'function') {
+          const observe$ = options.observe(this.context.app);
```
markvincze: Is the name ending with `$` a pattern?

markvincze (approved):

src/components/mapToProps.js
fahad19: yep, Observables (and streams) variable names usually end with a `$`.

PR #40, opened 2016-11-03
State as an Observable

fahad19:
## What's done:

* New method introduced in app instance called `getState$()`, which returns an Observable for the app's state
* `mapToProps` now uses the state Observable
* `react-redux` is no longer needed for this, and have been removed

Other changes:

* Some methods/variables have been renamed (still kept in code, but deprecated with warnings) to end with `$`, meaning they are Observables
* More tests added to improve coverage
* Report per-file coverage report in Terminal when running `$ npm test`

### Note:

* The PR is still fully **backwards compatible**, and no breaking changes have been added
* There are warnings for methods that have been deprecrated
* Those methods will be removed from the codebase when we release the next breaking semver in future

## API additions:

In `App` instance:

* `getState$()`: Returns state as an observable
* `dispatch()` Dispatches an Action to Store
* `observeWidgets$()`: Returns observable for Widgets

## Usage example:

You are not expected to use the `app.getState$()` method directly, since `mapToProps` takes care of it for you. But the usage example is as follows:

```js
const state$ = app.getState$();

const subscription = state$.subscribe((state) {
  // will be triggered as many times the state changes
  console.log('state changed to:', state);
});

subscription.unsubscribe();
```


alexmiranda:
This affects `frint-test`, so please hold on the merge until I create my PR to `frint-test` repository.

AlexDudar:
@fahad19 you said that thisPR is still fully backwards compatible, but you also removed some stuff in the documentation..of course you can do it, and that's fine, but what if in future you will have some braking changes and we can't move to new version, it would be nice to have old documentation as well.

fahad19:
@AlexDudar: the removed stuff in docs were incorrect, that's why the message says `remove docs for mapToProps's dispatch key, which was incorrect.`.

PR #50, opened 2016-11-23
Adds greenkeeper ignore and updates .gitignore

mAiNiNfEcTiOn:
# What does this PR do:
* Adds common patterns for 'macOS', 'OSX', 'VisualStudioCode', 'SublimeText' and 'WebStorm' to the `.gitignore`.
* Adds a section _greenkeeper_ to the `package.json` in order to specify packages to ignore (we don't want to upgrade `react` and `react-dom` for now)

  - Diffs

N/A

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,32 +1,10 @@
 {
-  "name": "frint",
-  "version": "0.9.0",
-  "description": "Framework for building front-end apps",
-  "main": "./lib/index.js",
-  "scripts": {
-    "build": "webpack",
-    "transpile": "babel src --out-dir lib",
-    "lint": "eslint --color '{src,test}/**/*.js'",
-    "cover": "nyc --reporter=lcov --require babel-register mocha && npm run cover:report",
-    "cover:report": "nyc report",
-    "test": "npm run cover",
-    "coverage:coveralls": "cat ./coverage/lcov.info | coveralls",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -81,5 +55,34 @@
...
     "sinon-chai": "^2.8.0"
-  }
+  },
+  "greenkeeper": {
+    "ignore": ["react", "react-dom"]
+  },
+  "homepage": "https://github.com/Travix-International/frint#readme",
+  "license": "MIT",
+  "main": "./lib/index.js",
+  "name": "frint",
```
fahad19: a `package.json` file not starting with the `name` key looks really weird. is it alright to have an exception for not ordering the keys alphabetically in `package.json`?

fahad19:
noticed in one of the PRs by @asci, that he is copying the contents of gitignore to `.npmignore` too. otherwise everything gets published to npm (including our bundled _book directory). shall we update the `npmignore` file here too?

package.json
mAiNiNfEcTiOn: -.-' Of course we can, but IMHO, that just doesn't matter... Automated stuff couldn't care less and if you trust in the package.json's name attribute to know what's the project you're on, then ....

Still, reverting to non sorted.

PR #51, opened 2016-12-07
Uses alexjs's ability to receive globs in order to parse other extensions

mAiNiNfEcTiOn:
# What does this PR do:
* Changes `alex:code` script on `package.json` to pass a specific glob to `alex` instead of doing a `find`. `alex` already ignores `node_modules` so no need to ignore them either.

  - Diffs

N/A


fahad19:
💖

PR #56, opened 2016-12-12
Non-blocking Regions for Widgets sharing state

fahad19:
## Problem

Let's take a look at this scenario:

* Widget `Foo` (primary Widget)
* Widget `Bar` (secondary, reads shared state from `Foo`)

The way `Region` is implemented, Widget `Bar` won't be rendered until `Foo` has been successfully instantiated in the browser.

This is all fine, until `Bar` needs to read state from `Foo`.

This causes problem, because sometimes needs arise where both Widgets need to read state from each other.

## Solution

* `Region` has been made non-blocking, meaning it would render both the Widgets (doesn't matter if one is not loaded at any given point in time)
* If any shared state is not available, it will expose it as an empty object for the time being
* Once the shared state is available (i.e., after the Widget is loaded), it will map its state accordingly
* Currently, it looks for whether the requested Widget has been loaded or not every 100ms, and once it is available, it kills the Observable subscription (avoiding any memory leak)

## Other things done

* Upgraded `rxjs` to latest stable 5.x
* Example applications in `examples` directory
* Tests coverage

## Preview

* `Foo` has counter
* `Bar` has color
* They both share their state with each other

You can run the example yourself, by running:

```
$ npm install

$ cd examples/multiple-widgets
$ make serve
```

screen shot 2016-12-12 at 17 37 52


fahad19:
(will take care of the coverage soon)

PR #61, opened 2016-12-14
Check if `window` exists before setting Root app

fahad19:
This check is needed to use `frint` in server-side, without involving any `jsdom` magic in globals.

### DIFF
--- a/src/createApp.js
+++ b/src/createApp.js
@@ -53,7 +53,10 @@ class BaseApp {
     // widgets
     this.widgetsByRegion = {};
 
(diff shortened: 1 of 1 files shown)


src/createApp.js
```diff
@@ -53,7 +53,10 @@ class BaseApp {
     // widgets
     this.widgetsByRegion = {};
 
-    if (typeof window.app !== 'undefined') {
+    if (
```
mAiNiNfEcTiOn: Yeah.. verifying for `process` would also be a thing :P

PR #63, opened 2016-12-15
Hyperscript for JSX

fahad19:
## Problem

We are still dependent on JSX being transpiled into code like `React.createElement(...)`, which makes it a requirement for us to expose `React` globally.

In the browser we are not facing any issues, because React is available globally already as `window.React`.

But in the server-side, it's a problem unless you set `global.React = require('react')`.

## Solution

Instead of depending on React globally, we are now exposing an `h` function, which we can import wherever we use JSX.

And it comes from `frint`.

`h` is the replacement of direct usage of `React.createElement`.

### Component

The `h` function needs to exist in the scope, wherever we use JSX.

```js
import { createComponent, h } from 'frint';

export default createComponent({
  render() {
    return (
      <p>Hello World</p>
    );
  }
});
```

And it gets transpiled to:

```js
import { createComponent, h } from 'frint';

export default createComponent({
  render() {
    return h(
      'p',
      null,
      'Hello World'
    );
  }
});
```

### Babel

File-level, with comment:

```js
/** @jsx h */
import { createComponent, h } from 'frint';

export default createComponent({
  render() {
    return <p>Hello World</p>
  }
});
```

If you don't want to set it in every file as a comment, you can set it globally in your `.babelrc` file:

```json
{
  "presets": [
    "travix"
  ],
  "plugins": [
    ["transform-react-jsx", {
      "pragma": "h"
    }]
  ]
}
```

### ESLint

Your `.eslintrc` file can look like:

```json
{
  "extends": [
    "travix"
  ],
  "parser": "babel-eslint",
  "settings": {
    "react": {
      "pragma": "h"
    }
  }
}
```

---

**Note**: Not a breaking change.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,5 +1,10 @@
 {
   "presets": [
     "travix"
+  ],
+  "plugins": [
+    ["transform-react-jsx", {
+      "pragma": "h"
+    }]
   ]
 }
--- a/.eslintrc
+++ b/.eslintrc
@@ -2,5 +2,10 @@
   "extends": [
     "travix"
   ],
-  "parser": "babel-eslint"
+  "parser": "babel-eslint",
+  "settings": {
+    "react": {
+      "pragma": "h"
+    }
+  }
 }
--- a/SUMMARY.md
+++ b/SUMMARY.md
@@ -27,6 +27,7 @@
 * [API Reference](/docs/api/README.md)
   * [createApp](/docs/api/createApp.md)
   * [createComponent](/docs/api/createComponent.md)
+  * [h](/docs/api/h.md)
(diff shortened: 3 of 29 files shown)


markvincze:
Maybe I misunderstand how this works, but can't we make it so that `h` is automatically imported whenever we import `createComponent`?

fahad19:
> Maybe I misunderstand how this works, but can't we make it so that h is automatically imported whenever we import createComponent?

@markvincze: unfortunately not. that's the same reason why `React` is imported in regular react apps too. even though only `React.createClass` is needed for creating the Components. Transpiled code can then pick up `React.createElement` because imported `React` is in the scope.

PR #71, opened 2016-12-28
Store management with `createStore`

fahad19:
## What's done

* New `createStore` function
  * Used internally for handling Stores for App instances
* Dropped `redux`
* Fully **backwards compatible** with deprecated methods
* Full test coverage
* Minified/gzipped dist file down to `~7kB` from `~13kB`

## Preview

Both examples in the repo working, and here's a screenshot for Multiple Widgets:

screen shot 2016-12-28 at 15 52 08


src/createStore.js
```diff
@@ -0,0 +1,127 @@
...
+      enableLogger: true,
+      console: console,
+      ...options,
+    };
+
+    this.internalState$ = new BehaviorSubject(this.options.initialState)
+      .scan((previousState, action) => {
+        let updatedState;
+        const d = new Date();
+        const prettyDate = `${d.getHours()}:${d.getMinutes()}:${d.getSeconds()}.${d.getMilliseconds()}`;
```
mAiNiNfEcTiOn: Remember, you don't have zero-padded integers.

src/createStore.js
```diff
@@ -0,0 +1,127 @@
...
+          if (action && action.type) {
+            this.options.console.error(`Error processing @ ${prettyDate} ${action.type}:`);
+          }
+          this.options.console.error(error);
+
+          return previousState;
+        }
+
+        // logger in non-production mode only
+        if (process.env.NODE_ENV !== 'production') {
```
mAiNiNfEcTiOn: Perhaps this logic should go to the logger itself, or better yet, whenever you create an instance of the logger you _could_ specify the environment(s) in which it should work?

Why? Because of this: https://github.com/nodejs/node/issues/3104

Another way would be to cache that value at the beginning of the module. :)

src/createStore.js
```diff
@@ -0,0 +1,127 @@
...
+    return this.exposedState$;
+  }
+
+  getState = () => {
+    this.options.console.warn('[DEPRECATED] `Store.getState` has been deprecated, and kept for consistency purpose only with v0.x');
+
+    return this.cachedState;
+  }
+
+  dispatch = (action) => {
```
mAiNiNfEcTiOn: Is this `action` the `String` (e.g. `'MY_ACTION'`) or the `Object` passed with the type?

markvincze:
The couple of deleted test files, like appendAction, and async: why aren't those needed anymore?

And just to make sure I understand correctly: instead of depending on `redux`, you implemented a small part of its functionality in `createStore.js`?

fahad19:
@markvincze: they are deleted because they were middlewares for Redux, not they are supported directly via `createStore` function, in the form of passed `options`.

See these `options` in `createStore.js`:

* `thunkArgument`: for deleted `async.js` middleware
* `appendAction`: for deleted `appendAction.js` middleware
* `enableLogger`: for deleted `logger.js` middleware

src/createStore.js
fahad19: @mAiNiNfEcTiOn: I kept it like it was working before. The main reason is because of our production bundles.

When `NODE_ENV === production`, the full code-block inside this `if` would not end up in the bundle processed by Webpack (in our case, it is `vendors.js` file).

If we want to check it in runtime instead, it means the logger specific code would have to be in the bundle all the time. Do we want that in this PR?

Please correct me if you understand your comment wrong.

src/createStore.js
fahad19: @mAiNiNfEcTiOn: sorry for the confusion. I should add some docblocks soon.

`Store.dispatch` can accept:

* a payload, like: `{type: 'DO_SOMETHING'}`, or
* a function, like:
   ```js
   function (dispatch, getState, { app }) {
     aTimeConsumingPromise()
       .then(function () {
         dispatch({ type: 'DO_SOMETHING' });
       });
   }
   ```

It maintains the sync/async actions, just like how we support them now.

src/createStore.js
mAiNiNfEcTiOn: Ok, clear then.

src/createStore.js
mAiNiNfEcTiOn: ;) Sure... good good. That's ok then :) 👍

src/createStore.js
```diff
@@ -18,7 +18,15 @@ class BaseStore {
       .scan((previousState, action) => {
         let updatedState;
         const d = new Date();
-        const prettyDate = `${d.getHours()}:${d.getMinutes()}:${d.getSeconds()}.${d.getMilliseconds()}`;
+        const prettyDate = [
+          _.padStart(d.getHours(), 2, 0),
```
mAiNiNfEcTiOn: -.-' One liners...:

`((str) => '0'.repeat(Math.abs(str.length - 2)) + str)(d.getHours())`...

But what do I know? 😄

PR #73, opened 2017-01-02
CNAME for frint.js.org

fahad19:
Once the `CNAME` file gets deployed to GitHub Pages, we can request for a `frint.js.org` domain name via https://dns.js.org/

### DIFF
--- a/CNAME
+++ b/CNAME
@@ -1,0 +1,1 @@
+frint.js.org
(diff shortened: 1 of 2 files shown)


fahad19:
Here is the other PR. should go live within 24 hours once merged:

https://github.com/js-org/dns.js.org/pull/987

Issue #74, opened 2017-01-03
`providers` for combining Services and Factories


fahad19:
Instead of developers explicitly requesting for a particular Service of Factory, let them request for the instance just by their name (without having to know whether it is a Service or a Factory).

Example API (doesn't mean it will be the final API):

```js
import FooService from './services/Foo';
import BarFactory from './factories/Bar';

const App = createApp({
  providers: {
    foo: FooService,
    bar: BarFactory
  }
});

const app = new App();

const fooService = app.get('foo'); // returns service
const barFactory = app.get('bar'); // returns factory

// same way, mapToProps() can also have `providers` key:
const ContainerComponent = mapToProps({
  providers: {
    foo: 'foo',
    bar: 'bar',
  }
})(RootComponent);
```

fahad19:
If we happen to have these types of classes:

* Service: Singleton, returns same instance every time from anywhere
* Scoped Service: Singleton, but scoped by the App that requested it
* Factory: Always returns a new instance, scoped by the App that requested it

We can possibly define them all with just the `createService` function:

```js
const FooService = createService(...);
const BarFactory = createService(...);
```

But when defining them in the App class, they can be:

```js
const App = createApp({
  providers: [
    foo: { provide: FooService }, // singleton
    scopedFoo: { provide: FooService, scoped: true }, // scoped Singleton
    bar: { provide: BarFactory, factory: true }, // returns scoped factory
  ]
});
```

fahad19:
the more I think about, the more I feel `providers` can be used for injecting everything, even models and global variables, like this for example:

```js
const App = createApp({
  providers: {
    // services, returning instances from classes
    foo: { provide: FooService }, // singleton
    scopedFoo: { provide: FooService, scoped: true }, // scoped Singleton
    bar: { provide: BarFactory, factory: true }, // returns scoped factory
  }
});

const app = new App({
  providers: {
    // use value directly
    window: { provide: window, useValue: true }, 
    document: { provide: window.document, useValue: true },
    
    // models, using value directly
    configuration: { provide: new ConfigurationModel({ key: 'value' }), useValue: true },
    affiliate: { provide: new AffiliateModel({ brand: 'X' }), useValue: true },
  }
});

app.get('window'); // returns global `window` object
app.get('document'); // returns global `document` object

app.get('configuration'); // returns same instance of ConfigurationModel every time (singleton)
app.get('affiliate'); // returns same instance of Affiliate model

[...]

PR #76, opened 2017-01-10
Combine services and factories

alexmiranda:
This is related to [issue 74](https://github.com/Travix-International/frint/issues/74).

For now you can see that little unit test code has been changed.  I will later add extra unit tests that explain `singletonServices` v `appScopedServices` (aka. factories).

The only breaking change is that services or factories depending on an app instance will have to inject the `app` via `initialize` method (which used to be automatic).

The file `_mergeServicesAndFactories.js` is supposed to be temporary (thus the weird naming convention).

### DIFF
--- a/src/_mergeServicesAndFactories.js
+++ b/src/_mergeServicesAndFactories.js
@@ -1,0 +1,12 @@
+// TODO: remove this when factories are removed
+export default function mergeServicesAndFactories(options) {
+  const { services = {}, factories = {} } = options;
+  const collidingFactoryNames = Object.keys(factories)
+    .filter(factory => factory in services && factories[factory] !== services[factory]);
+  if (collidingFactoryNames.length) {
+    throw new Error(`Colliding factory names: ${collidingFactoryNames.join(', ')}`);
+  }
+
+  Object.assign(services, factories);
+  delete options.factories;
+}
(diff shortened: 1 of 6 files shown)


src/_mergeServicesAndFactories.js
```diff
@@ -0,0 +1,12 @@
+// TODO: remove this when factories are removed
+export default function mergeServicesAndFactories(options) {
+  const { services = {}, factories = {} } = options;
+  const collidingFactoryNames = Object.keys(factories)
+    .filter(factory => factory in services && factories[factory] !== services[factory]);
+  if (collidingFactoryNames.length) {
+    throw new Error(`Colliding factory names: ${collidingFactoryNames.join(', ')}`);
```
mAiNiNfEcTiOn: At what level is this error is being triggered?

If an app triggers this error, would it block the load of the rest of the apps or only of that specific app?

src/createFactory.js
```diff
@@ -1,12 +1,7 @@
...
 
-export default function createFactory(extend = {}) {
-  const Service = createService(extend);
-
-  class Factory extends Service {
-
-  }
-
-  return Factory;
+export default function createFactory(...args) {
```
mAiNiNfEcTiOn: wouldn't it be better to do:

`export default createService;`

?

fahad19:
I believe this PR can be closed without merging, since it is done in the `next` branch?

alexmiranda:
This has been addressed in `frint-next` and already merged into `master`.  I'm closing this now.

PR #88, opened 2017-02-03
Improve model functionality

spzm:
There are few improvements are proposed in the PR:

1. Always create an object for attributes field when creating a Model. It will prevent situation when the single value or undefined can be passed to it.

2. Introduce the get method as a common approach to retrieve a value from attributes field. The get method support complex keys and actually uses lodash get.

### DIFF
--- a/src/Model.js
+++ b/src/Model.js
@@ -2,7 +2,16 @@ import _ from 'lodash';
 
 export default class Model {
   constructor(attributes) {
-    this.attributes = attributes;
+    this.attributes = {};
+    Object.assign(this.attributes, attributes);
+  }
+
+  get(key) {
+    if (typeof key !== 'string'
+      || !Object.hasOwnProperty.call(this.attributes, key)) {
+      return undefined;
+    }
+    return this.attributes[key];
   }
 
   toJS() {
--- a/src/Model.js
+++ b/src/Model.js
@@ -2,8 +2,7 @@ import _ from 'lodash';
 
 export default class Model {
   constructor(attributes) {
-    this.attributes = {};
-    Object.assign(this.attributes, attributes);
+    this.attributes = Object.assign({}, attributes);
   }
 
   get(key) {
--- a/src/Model.js
+++ b/src/Model.js
@@ -7,11 +7,9 @@ export default class Model {
   }
 
   get(key) {
-    if (typeof key !== 'string'
-      || !Object.hasOwnProperty.call(this.attributes, key)) {
-      return undefined;
-    }
-    return this.attributes[key];
+    if (typeof key !== 'string') return undefined;
+
+    return _.get(this.attributes, key);
   }
 
   toJS() {
--- a/test/Model.spec.js
+++ b/test/Model.spec.js
@@ -23,4 +23,14 @@ describe('Model', () => {
     expect('toJS' in myModelInstance).to.be.equal(true);
     expect(myModelInstance.toJS()).to.be.deep.equal(myAttributes);
   });
+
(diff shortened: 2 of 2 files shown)


src/Model.js
```diff
@@ -2,7 +2,14 @@ import _ from 'lodash';
...
   constructor(attributes) {
-    this.attributes = attributes;
+    this.attributes = {};
+    Object.assign(this.attributes, attributes);
+  }
+
+  get(key) {
+    if (typeof key !== 'string') return undefined;
+
+    return _.get(this.attributes, key);
```
mAiNiNfEcTiOn:
```js
// on top of the field
const FIELD_SEPARATOR = '.';

// in the model
  get(key) {
    if ((typeof key === 'string') && (key.length)) {
      return key.split(FIELD_SEPARATOR).reduce((ret, field) => {
        if ((typeof ret === 'object') && (ret !== null)) {
          return ret[field];
        }
        return ret;
      }, this.attributes);
    }
    return undefined;
  }
```

(Ignore this message, I just hate lodash usage for such small behaviour 😝 )

src/Model.js
```diff
@@ -2,7 +2,14 @@ import _ from 'lodash';
 
 export default class Model {
   constructor(attributes) {
-    this.attributes = attributes;
+    this.attributes = {};
```
spzm: thanks, I have update. Regarding references, the alternative solution is cloneDeep but it can cause the performance. I think it is matter of discuss.

fahad19:
Can you please update the docs too?

Mainly:

* Update `docs/Api/createModel.md`
* May be create a new `docs/Api/Model.md` file too since this is an instance method

alexmiranda:
Is this really necessary?  I mean, why would we be concerned with creating nesting and structures into models whilst most of its use it to access static readonly data provided by parent/root application?  I understand that for representing some attributes it may be reasonable that you group some attributes e.g. address, pricing, etc; however, this is almost never the case with the _fireball_ applications that we have, i.e. models support configuration and settings injected into components.  If we introduce models to carry data from booking flow (more structured, dynamic data), for instance, why would you want to complicate access pattern if you can very well use a well known json object/schema?

This is not to say I'm against this in any way.  If you have any specific use case that would really benefit from this, bring it up.  I'd be glad to hear.  Other than that, YAGNI.

mAiNiNfEcTiOn:
@alexmiranda Although the 'nested get' is not priority, we should definitely have into consideration the fact that the attributes are being defined by reference:

```js
// MyModel.js
import { createModel } from 'frint';
export default createModel({
  set(key, value) {
    this.attributes[key] = value;
  }
});

// Dependent
import MyModel from './MyModel';

const obj = {
  prop: 'value',
  nested: {
    prop: 'nested value',
  },
};

const model = new MyModel(obj);
model.set('prop', 'new value');

console.log(obj.prop === 'new value'); // true
```

spzm:
My main goal was to resolve the issue when you make new MyModel(undefined), then empty attributes objects will be created. Then in initialize method of createModel I have to check - is it object or not. Also the case when the argument is any other primitive type.

The getter idea become useful when we get the values from the model. Also it help to understand that all the model data have to be stored in attributes. The getter also can be useful a lot inside model methods.

Also one of my idea was to freeze model object and attributes after creation. Then it will more read only. In my view, model was something that we creates ones when data comes. And then using it as read only data. It how we used it in +/- 3 days widget. I'm curious about real examples when we need to change the model after creation. 

Model changes can make the structure of model data quiet different. Then it can make a lot of issues when I using it. It would be good if the model will be something that I can trust (structure and types I mean).

PR #91, opened 2017-02-27
Extract lifecycle hooks to react package

alexmiranda:


fahad19:
these can now also be removed from `./src/core/test/App.spec.js`: https://github.com/Travix-International/frint/blob/d6f03f7715710f64af2e8f75f3aa7aa7c11da432/src/core/test/App.spec.js#L224

src/react/extendApp.js
```diff
@@ -12,4 +12,15 @@ export default function extendApp(App) {
       </Provider>
     );
   };
+
+  ['beforeMount', 'afterMount', 'beforeUnmount'].forEach((hookName) => {
+    App.prototype[hookName] = function lifecycleHook(...args) {
```
fahad19: just to be clear, here are we doing this? brain had a hard time parsing the code 😅 

* extract hook by name
* if not a function, default to noop
* run the function against `app`'s context with the given arguments, and return its output
* if successfully run, set the function as a method in the `app` instance

fahad19:
Thanks for doing this! much cleaner now 💖

PR #109, opened 2017-03-09
App lifecycle (destroy)

fahad19:
## What's done

* When multi-instance Widgets are unmounted:
  * The instances are destroyed
  * And the root app also clears them from internal registry

This is done because multi-instance Widgets are directly dependent on repeating Regions. If Regions are unmounted, they become redundant and add to memory leakage.

## Usage

Nothing extra needs to be done from a developer's point of view, but a new `beforeDestroy` callback is now available:

```js
import { createApp } from 'frint';

const App = createApp({
  name: 'MyApp',

  initialize: function () {
    // called when the App is instantiated
  },

  beforeDestroy: function () {
    // called when App is about to be destroyed
  }
});
```

### DIFF
--- a/packages/frint-react/src/components/Region.js
+++ b/packages/frint-react/src/components/Region.js
@@ -132,7 +132,7 @@ export default React.createClass({
 
     if (this.rootApp) {
       this.state.listForRendering
-        .filter(item => item.multi === true)
+        .filter(item => item.multi)
         .forEach((item) => {
           this.rootApp.destroyWidget(
             item.name,
--- a/packages/frint-react/src/components/Region.js
+++ b/packages/frint-react/src/components/Region.js
@@ -60,6 +60,7 @@ export default React.createClass({
           this.state.list.forEach((item) => {
             const widgetName = item.name;
             const widgetWeight = item.weight;
+            const widgetMulti = item.multi;
             const existsInState = this.state.listForRendering.some((w) => {
               return w.name === widgetName;
             });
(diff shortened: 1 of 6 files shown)


packages/frint-react/src/components/Region.js
```diff
@@ -127,7 +130,17 @@ export default React.createClass({
       this.subscription.unsubscribe();
     }
 
-    // @TODO: clear instances
+    if (this.rootApp) {
+      this.state.listForRendering
+        .filter(item => item.multi === true)
```
jackTheRipper: `multi` is a boolean, wouldn't it be the same to write `.filter(item => item.multi)` ?

packages/frint/src/App.js
```diff
@@ -294,6 +295,34 @@ App.prototype.instantiateWidget = function instantiateWidget(name, region = null
...
+
+  const w = this._widgetsCollection[index];
+  const key = makeInstanceKey(region, regionKey, w.multi);
+
+  if (typeof this._widgetsCollection[index].instances[key] === 'undefined') {
+    throw new Error(`No instance with key '${key}' found for widget with name '${name}'.`);
+  }
+
+  this._widgetsCollection[index].instances[key].beforeDestroy();
+  delete this._widgetsCollection[index].instances[key];
```
jackTheRipper: Shouldn't it be done in `diyai` as well ?

fahad19: like we just discussed: App dies, container dies.

PR #113, opened 2017-03-12
REPL

fahad19:
## What's done

* Live code editing and previewing
* [Ace](https://ace.c9.io/#nav=about) editor used
* Libraries are loaded in the browser, and made available as these variables:
  * `_`
  * `Rx`
  * `React`
  * `ReactDOM`
  * `Frint`
  * `FrintModel`
  * `FrintStore`
  * `FrintReact`
  * `Babel`

## Before merge

* [x] Needs some CSS fixing

## Build

Expects the dist files to be already available in the disk for copying to generated site. The dists can be generated as follows:

```
$ npm run bootstrap
$ npm run dist
```

## Ideas for further extension (in future)

* Some pre-coded snippets to select from a drop down, and try them out
* Save user-written code in localStorage, and make it available in next page load
* [x] Generate URL for sharing (code can be embedded in the URL itself)

## Preview

screen shot 2017-03-12 at 20 56 35

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -47,10 +47,7 @@ site-build:
 
 	cp -rf ./site/assets/img ./_site/img
 
-	cp -rf ./packages/frint/dist ./_site/js
-	cp -rf ./packages/frint-model/dist ./_site/js
-	cp -rf ./packages/frint-store/dist ./_site/js
-	cp -rf ./packages/frint-react/dist ./_site/js
+	cp -rf ./packages/frint*/dist ./_site/js
 
 site-watch:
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -47,6 +47,11 @@ site-build:
 
 	cp -rf ./site/assets/img ./_site/img
 
+	cp -rf ./packages/frint/dist ./_site/js
```
jackTheRipper: You could shortcut it by using `cp -rf ./packages/frint*/dist ./_site/js`

PR #132, opened 2017-03-15
Coveralls integration for monorepo

fahad19:
## What's done

Combine coverage reports of multiple packages into one single lcov file, and then send it to coveralls.


fahad19:
Lowered the threshold to 97% from 100% on coveralls.io settings.

PR #152, opened 2017-03-25
App hierarchy

fahad19:
## What's done

* Parent/Child relationship between Apps is now fully hierarchical
  * We currently only have `Root App ==> Widget` hierarchy
  * With this PR, we now have `Root App ==> As ==> Many ==> Apps ==> Widget`
* More tests written
* Docs provided
* Does not change the behaviour of 1.x in any way, and in fact makes it purer

## Future

This PR now makes way for things like: 

* Routing system with nested views
* Concept of Shell App that we discussed before (for PWA), which can closely work with Service Workers

### DIFF
--- a/packages/frint-compat/src/createApp.spec.js
+++ b/packages/frint-compat/src/createApp.spec.js
@@ -103,7 +103,9 @@ describe('frint-compat › createApp', function () {
       }
     });
 
-    const widget = new WidgetApp();
+    const widget = new WidgetApp({
+      parentApp: window.app,
+    });
 
     expect(widget.getModel('baz').getName()).to.equal('Baz');
   });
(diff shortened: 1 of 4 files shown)


packages/frint-compat/src/createApp.spec.js
```diff
@@ -317,11 +323,12 @@ describe('frint-compat › createApp', function () {
...
 
-  it('throws error if setRegion is called without a root app', function () {
-    const widget = new WidgetApp();
+  // NOTE: setRegion will never be called in v1.x
+  // it('throws error if setRegion is called without a root app', function () {
+  //   const widget = new WidgetApp();
 
-    expect(() => widget.setRegion('sidebar')).to.throw(/No root app instance available/);
-  });
+  //   expect(() => widget.setRegion('sidebar')).to.throw(/No root app instance available/);
```
AlexDudar: why you didn't remove this?

fahad19: I moved all the tests of frint v0.x to `frint-compat` and kept them unchanged to make sure we can depend on `frint-compat` reliably. 

If I really had to remove any individual test, I added a `NOTE` on top of the commented block explaining why it's not needed any more. 

Just to keep a reference.

PR #155, opened 2017-03-29
Add search functionality with Algolia

markvincze:
This is what it looks like, on the home page

image

image

On other pages:

image

image

Inside the search popup the little underlines are blue, but I couldn't change them with the recommended styles (https://community.algolia.com/docsearch/documentation), so probably we'd have to dig deeper to adjust that.
Please tell if anything should be done differently, and feel free to jump in and change the styles if you'd like to change the looks.

### DIFF
--- a/site/partials/navLinks.html
+++ b/site/partials/navLinks.html
@@ -20,3 +20,14 @@
     About
   </a>
 </div>
+
+<div class="nav-right">
+  <div class="field nav-item">
+    <p class="control has-icon">
+      <input id="doc-search-text-box" class="input" type="text" placeholder="Search...">
+      <span class="icon is-small">
+        <i class="fa fa-search"></i>
+      </span>
+    </p>
+  </div>
+</div>
--- a/site/partials/assets.html
+++ b/site/partials/assets.html
@@ -4,11 +4,22 @@
 
 <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/4.7.0/css/font-awesome.min.css" />
 <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/prism/1.6.0/themes/prism-solarizedlight.min.css" />
+<link rel="stylesheet" href="https://cdn.jsdelivr.net/docsearch.js/2/docsearch.min.css" />
 <link rel="stylesheet" href="/css/site.css" />
 
 <script src="https://cdnjs.cloudflare.com/ajax/libs/jquery/3.1.1/jquery.min.js"></script>
 <script src="https://cdnjs.cloudflare.com/ajax/libs/prism/1.6.0/prism.min.js"></script>
 <script src="/js/site.js"></script>
+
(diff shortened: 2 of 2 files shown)


site/assets/css/header.scss
```diff
@@ -10,6 +10,49 @@
...
+  border: none;
+  font-size: 12pt;
+  background-color: #fafafa;
+  float: left;
+  padding-left: 40px;
+  -webkit-border-radius: 5px;
+  -moz-border-radius: 5px;
+  border-radius: 5px;
+  outline: none;
+}
```
markvincze: I'm not happy that I needed to add so many styles, so tell if you have any suggestions how to make it simpler.

site/assets/css/header.scss
```diff
@@ -10,6 +10,49 @@
   }
 }
 
+.nav-left {
+  overflow: visible;
+  overflow-x: visible;
+  overflow-y: visible;
```
markvincze: These were needed, otherwise the popup wasn't visible at all under the nav header.

fahad19:
awesome stuff, @markvincze! super impressed! 💖 

I would be very curious to see how the searchbox looks like if aligned on the right side.

You may find these docs of Bulma very handy: http://bulma.io/documentation/components/nav/

markvincze:
@fahad19,
This is what it's like on the home page:

image

And on another page:

image

Which one would you like? (I think I prefer this version, when it's on the right)

markvincze:
Ah, Bulma has built-in styles for the textboxes, I'll look into that and try to simplify the css.

markvincze:
Ah, cool, I changed it to use the built-in Bulma form styles, so all the horrible css is gone (and it looks a bit better ^^)

site/partials/navLinks.html
```diff
@@ -20,3 +20,14 @@
     About
   </a>
 </div>
+
+<div class="nav-right">
+  <div class="field nav-item">
+    <p class="control has-icon">
+      <input id="docSearchTextbox" class="input" type="text" placeholder="Search...">
```
fahad19: usually we write element IDs with hyphens, like `doc-search-text-box`.

site/partials/navLinks.html
```diff
@@ -20,3 +20,14 @@
...
+      <span class="icon is-small">
+        <i class="fa fa-search"></i>
+      </span>
+    </p>
+  </div>
+</div>
```
fahad19: using EditorConfig would take care of adding new line automatically.

fahad19:
💯 ✨ 💖

site/partials/navLinks.html
markvincze: You mean having a trailing empty line is good practice?

site/partials/navLinks.html
markvincze: kebab-casing ^^

site/partials/navLinks.html
markvincze: Actually it was already set up in VSCode, but I probably manually removed the line

site/partials/navLinks.html
fahad19: I always get confused among kebab, camel, pascal, etc :P

PR #166, opened 2017-04-20
Make models streamable

fahad19:
## What's done

Models now support Observables, for streaming changes as they happen. This would be of very high impact for dealing with forms in observed React components.

## Usage

Import:

```js
import { createModel } from 'frint-model';
```

Create Model class:

```js
const Shirt = createModel({});
```

Building a new model instance:

```js
const shirt = new Shirt({
  color: 'blue',
  size: 'medium',
});
```

The model instance can also be observed for changes:

```js
shirt.get$().subscribe(function (shirtAttributes) {
  // triggered when the model had any change
});

shirt.get$('color').subscribe(function (color) {
  // triggered when the model's `color` key changes
});
```

### DIFF
--- a/packages/frint-model/src/Model.js
+++ b/packages/frint-model/src/Model.js
@@ -1,15 +1,44 @@
 import _ from 'lodash';
+import { BehaviorSubject } from 'rxjs';
 
 function Model(attributes) {
   this.attributes = Object.assign({}, attributes);
+  this.$ = null;
 }
 
-Model.prototype.get = function get(key) {
+function getFromAttributes(attributes, key) {
+  if (typeof key === 'undefined') {
+    return attributes;
+  }
+
   if (typeof key !== 'string') {
     return undefined;
   }
 
-  return _.get(this.attributes, key);
+  return _.get(attributes, key);
+}
+
+Model.prototype.get = function get(key) {
+  return getFromAttributes(this.attributes, key);
+};
+
+Model.prototype.set = function set(key, value) {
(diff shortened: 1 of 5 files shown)


packages/frint-model/src/createModel.spec.js
```diff
@@ -56,4 +56,44 @@ describe('frint-model › createModel', () => {
...
+    it('does not start observing until needed', function () {
+      const person = new Person({ name: 'Rowena Revenclaw' });
+      expect(person.$).to.equal(null);
+    });
+
+    it('streams all attributes', function (done) {
+      const person = new Person({ name: 'Helga Hufflepuff' });
+      person.get$().subscribe(function (personAttributes) {
+        expect(personAttributes).to.deep.equal({
+          name: 'Helga Hufflepuff',
```
markvincze: ❤️

packages/frint-model/src/Model.js
```diff
@@ -1,15 +1,44 @@
 import _ from 'lodash';
+import { BehaviorSubject } from 'rxjs';
 
 function Model(attributes) {
   this.attributes = Object.assign({}, attributes);
+  this.$ = null;
 }
 
-Model.prototype.get = function get(key) {
+function getFromAttributes(attributes, key) {
```
fahad19: till now, we have always maintained the API documentation in package's own README whenever we made changes in code. but I agree jsdoc would be a better approach.

would really welcome PRs to speed that process up :)

Issue #146, opened 2017-03-21
Site: text selection color in code-blocks


fahad19:
Currently shows a dark color, which doesn't really go with rest of the site:

![screen shot 2017-03-21 at 13 13 20](https://cloud.githubusercontent.com/assets/20046/24146942/51ea4220-0e38-11e7-87ec-96b8a8e5cd59.png)

fahad19:
This tutorial would be handy: https://css-tricks.com/almanac/selectors/s/selection/

AlexDudar:
which one do you like:

<img width="684" alt="screen shot 2017-04-23 at 6 41 20 pm" src="https://cloud.githubusercontent.com/assets/3122478/25315408/b5caff06-2854-11e7-98a5-354caed50c50.png">

<img width="677" alt="screen shot 2017-04-23 at 6 41 33 pm" src="https://cloud.githubusercontent.com/assets/3122478/25315409/b84e2df2-2854-11e7-8393-6458e4ef7fd9.png">

fahad19:
second one looks great!

PR #168, opened 2017-04-23
Changed selection color (#146)

AlexDudar:
Adjusted selection color (#146)

screen shot 2017-04-23 at 7 01 29 pm


fahad19:
PR needs to pull in latest `master` before merging. otherwise 👍 

on a different note, I think it's time we need to start thinking about implementing `autoprefixer`: https://github.com/Travix-International/frint/blob/aecc1fd7cd02fe08e214278993b07f7592cb78b8/Makefile#L46

PR #171, opened 2017-04-25
TR-13266 Typings for frint

ghost:
**Do**

Add the typings for the different Frint packages.

Since `frint-compat` is maintained for our old applications I consider that it doesn't require the typings.


markvincze (approved):

markvincze:
💯

packages/frint-react/index.d.ts
```diff
@@ -0,0 +1,51 @@
+import { App } from 'frint';
+import { BehaviorSubject, Observale } from 'rxjs';
```
fahad19: `Observable`.

packages/frint-react/index.d.ts
```diff
@@ -0,0 +1,51 @@
...
+
+export class Streamer {
+  constructor(...args: any[]);
+
+  set(value: any, ...args: any[]) : Streamer;
```
fahad19: any convention for spaces before and after `:`?

packages/frint/index.d.ts
```diff
@@ -0,0 +1,92 @@
...
+
+    getParentApps(): App[];
+
+    getProvider(name: string): any;
+
+    getProviders(): any[];
+
+    getRootApp(): App;
+
+    getWidgetInstance(name: string, region?: string, regionKey?: string): App;
```
fahad19: @alexmiranda: makes sense to include deprecated methods here?

jackTheRipper: Since our users can still use it I think yes.

packages/frint-react/index.d.ts
jackTheRipper: Not really. But we could stick to `whatever: isMyType` for everything

packages/frint-react/index.d.ts
```diff
@@ -0,0 +1,51 @@
...
+
+  emit() : void;
+
+  getProps$(): BehaviorSubject<any>;
+
+  getData$(): Observable<any>;
+}
+
+export class Provider extends React.Component {
+  propTypes: React.PropTypes = {
```
fahad19: should `React` be imported in this file first?

fahad19:
💖

packages/frint-react/index.d.ts
jackTheRipper: Yup, I had it in my global scope. Adding it explicitly

jackTheRipper:
Next step will be to create a full example.

For now:

screen shot 2017-04-25 at 16 06 20

And:

screen shot 2017-04-25 at 16 05 51


It looks nice :)

fahad19 (approved):

PR #177, opened 2017-05-02
Fix examples

ghost:
Fix some warnings in the Kitchensink
Fix the counter example

@fahad19 @alexmiranda @AlexDudar : could we remove the multiple widget? The functionality it demonstrates is implemented in the kitchensink

@fahad19 @alexmiranda : what do we do regarding the React specific warnings? (createClass)

### DIFF
--- a/examples/counter/vendors/vendor.js
+++ b/examples/counter/vendors/vendor.js
@@ -47,3 +47,13 @@
   }
   window['FrintReact'] = m;
 })();
+
+// frint-store
+(function () {
+  var m = require('frint-store');
+  if (typeof m.default !== 'undefined') {
+    window['FrintStore'] = m.default;
+    return;
+  }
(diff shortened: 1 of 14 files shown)


markvincze (approved):

examples/counter/vendors/vendor.js
```diff
@@ -47,3 +47,13 @@
   }
   window['FrintReact'] = m;
 })();
+
+// frint-store
+(function () {
```
fahad19: can this file have deps defined in a more readable way? like:

```js
window._ = require('lodash');
window.FrintStore = require('frint-store');
```

jackTheRipper: Sure but following what we discussed yesterday I will do it for the kitchensink.

fahad19 (approved):

PR #188, opened 2017-05-09
Docs: migration guide for components

fahad19:
Guide detailing how to use React directly.

## Preview

screen shot 2017-05-09 at 18 13 56


fahad19:
coveralls seems to be dead the whole day.

PR #189, opened 2017-05-10
Typescript example

jackTheRipper:
** Do **

Replace the `counter` widget by the `counter-ts` widget
Demonstrate that we can use Typescript with Frint
Fix few typings

Must be merged **AFTER** that one: https://github.com/Travix-International/travix-di/pull/3


packages/frint-react/package.json
```diff
@@ -28,6 +28,8 @@
     "frint"
   ],
   "dependencies": {
+    "@types/react": "^15.0.24",
```
fahad19: can we treat them as peerDependencies? like asking developers to install them themselves?

even `react` and `react-dom` are not installed by this package directly.

examples/counter-ts/core/reducers/counter.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export interface CounterInitialState {
+  value: number;
+}
+
+const INITIAL_STATE = {
+  value: 0,
+};
+
+export default function counter(state: CounterInitialState = INITIAL_STATE, action: { type: string }): any {
```
reaktivo: Instead of having constants defined in a separate file, it would be nice if the CounterAction would be exported, including the constants it allows:

```
export enum CounterActionType {
  DECREMENT_COUNTER,
  INCREMENT_COUNTER,
}
export interface CounterAction {
  type: CounterActionType
}
```

examples/counter-ts/core/reducers/counter.ts
```diff
@@ -0,0 +1,27 @@
...
+  switch (action.type) {
+    case INCREMENT_COUNTER:
+      return Object.assign({}, {
```
reaktivo: No need for `Object.assign` here, this is equivalent to `return { value: state.value + 1 }`

examples/counter-ts/core/reducers/counter.ts
```diff
@@ -0,0 +1,27 @@
...
+        value: state.value + 1,
+      });
+    case DECREMENT_COUNTER:
+      return Object.assign({}, {
```
reaktivo: Same here.

examples/counter-ts/tslint.json
```diff
@@ -0,0 +1,11 @@
+{
+  "extends": ["tslint:latest", "tslint-react"],
+  "rules": {
+    // override tslint-react rules here
```
reaktivo: Some IDE's might complain, since this comment is not valid json.

examples/counter-ts/webpack.config.ts
```diff
@@ -0,0 +1,29 @@
...
+        ],
+    },
+    output: {
+        filename: "bundle.js",
+        path: join(__dirname, "build"),
+    },
+    // Enable sourcemaps for debugging webpack"s output.
+    resolve: {
+        // Add ".ts" and ".tsx" as resolvable extensions.
+        extensions: [".webpack.js", ".web.js", ".ts", ".tsx", ".js"],
```
reaktivo: Seems like `.webpack.js` and `.web.js` would be handled by the `.js` resolver.

examples/counter-ts/webpack.config.ts
```diff
@@ -0,0 +1,29 @@
+import { join } from "path";
```
reaktivo: This file has different indentation than the rest of the files in this PR.

examples/counter-ts/core/components/Root.tsx
```diff
@@ -0,0 +1,61 @@
...
+import * as React from "react";
+
+import {
+  decrementCounter,
+  incrementCounter,
+} from "../actions/counter";
+
+export interface RootProps extends ProviderProps {
+  counter: number;
+  incrementCounter: () => void;
```
jackTheRipper: No it doesn't. It's synchronous actions

jackTheRipper: My bad :D Badly copy-pasted xD

examples/counter-ts/core/actions/counter.ts
```diff
@@ -0,0 +1,17 @@
...
+  type: CounterActionType;
+}
+
+export function incrementCounter() : CounterAction {
+  return {
+    type: CounterActionType.INCREMENT_COUNTER,
+  };
+}
+
+export function decrementCounter() {
```
reaktivo: Missing CounterAction

fahad19:
💖

examples/counter-ts/core/reducers/counter.ts
```diff
@@ -0,0 +1,21 @@
...
+}
+
+const INITIAL_STATE = {
+  value: 0,
+};
+
+export default function counter(state: CounterState = INITIAL_STATE, action: { type: CounterActionType }): CounterState {
+  switch (action.type) {
+    case CounterActionType.INCREMENT_COUNTER:
+      state.value += 1;
```
reaktivo: I thought we were working with the Redux mentality, state should not be mutated.

examples/counter-ts/core/reducers/counter.ts
```diff
@@ -0,0 +1,21 @@
...
+      break;
+    case CounterActionType.DECREMENT_COUNTER:
+      state.value -= 1;
```
reaktivo: Same here

examples/counter-ts/core/reducers/counter.ts
jackTheRipper: That's why I was creating a new object using `Object.assign` :)

PR #198, opened 2017-05-15
Examples: Use Webpack HTML plugin to build index.html, and fix npm script

asci:
## What
- Use `html-webpack-plugin` to build `index.html` page automatically
- fix for `npm run build` script


examples/multiple-apps/build/index.html
```diff
@@ -12,9 +12,5 @@
     <script src="https://cdnjs.cloudflare.com/ajax/libs/lodash.js/4.17.3/lodash.min.js"></script>
```
fahad19: if this file is generated by Webpack, let's delete it from the repo, and also ignore it from `.gitignore`.

fahad19:
(I am taking care of the ESLint fails on Travis-CI on a separate branch)

fahad19:
@asci: please update your branch from latest `master`, and the build error will go away.

examples/multiple-apps/build/index.html
asci: done

examples/multiple-apps/.gitignore
```diff
@@ -0,0 +1 @@
+build
```
fahad19: I think only `build/index.html` would be enough.

fahad19:
one problem of deleting the entire `build` directory is, Webpack bundling would fail with an error for new users.

keep a file like `build/js/.gitkeep` to avoid this issue.

jackTheRipper:
What about the other examples?

examples/multiple-apps/.gitignore
alexmiranda: Probably it makes more sense to exclude the `build` folder altogether, unless you plan to have anything non-generated in that folder.

asci:
The build time with CommonChunks plugin is too long. I reverted changes and provided another PR #200

PR #213, opened 2017-05-29
Upgrade all packages to React v15+

### LINKED ISSUE #205: Upgrade to React v15
And also see how we can keep things backwards compatible with React v0.14.

fahad19:
Closes #205 

## Before merge

* [x] ~Backwards compatibility in private projects~ We will go for a major release
* [x] Update docs
* [x] Release last 1.x version from `master`, and create a branch from it called `1.x`

## Breaking change

You are now required to install `prop-types` in your own project if you want to use `frint-react`.

The installation command with npm looks like this:

```
$ npm install --save react react-dom prop-types frint-react
```

## After merge

* [ ] Create migration guide based on the information above under `Breaking change`
* [ ] Release 2.x
* [ ] Update examples to use latest Frint and React versions


markvincze:
Does this mean that after this is released, we can only update frint in rwd if React is updated in rwd too?

fahad19:
@markvincze: yes, as one of the tasks mentions `Backwards compatibility in private projects`, which is our RWD.

fahad19:
now that I think about it, if we do something like this in our private projects using React v0.14, things should ideally work without any issues:

```js
// let's say in `vendors.js`
import React from 'react';
import createClass from 'react-create-class';
import PropTypes from 'prop-types';

window.React = React;
window.React.createClass = createClass;
window.React.PropTypes = PropTypes;
```

It doesn't look great, but should help us ease the process of the upgrade of React 0.14 to React v15.

It also helps that `frint` packages never install `react` and `react-dom` themselves, and are treated as peerDependencies. So it is up to the developer to install their preferred React versions in their projects.

/cc @mAiNiNfEcTiOn

packages/frint-compat/src/createComponent.js
```diff
@@ -21,30 +23,44 @@ export default function createComponent(options = {}) {
...
+    }
 
     componentWillMount() {
       if (typeof options.beforeMount === 'function') {
         return options.beforeMount.call(this);
       }
 
       return null;
-    },
+    }
```
alexmiranda: :-(

packages/frint-compat/src/render.spec.js
```diff
@@ -45,16 +43,17 @@ describe('frint-compat › render', function () {
...
-    expect(React.createClass)
-      .to.be.callCount(1)
-      .and.to.be.calledWith({
-        componentDidMount: sinon.match.func,
-        componentWillUnmount: sinon.match.func,
-        displayName: 'WrapperComponent',
-        render: sinon.match.func
-      });
-  });
+  // NOTE: commented because of upgrade to React v15
```
alexmiranda: Should maybe remove this comment block entirely?

fahad19: we should be removing the `frint-compat` package entirely instead 😛 

but I mostly keep old code lying there (usually with a `NOTE: ...` comment) so that compat can tell us if there is anything that's not compatible any more.

packages/frint-compat/src/createComponent.js
fahad19: it's not an object any more, that's why comma is removed.

PR #222, opened 2017-06-15
Fixes for `alex`.

fahad19:
## What's done

Added a few more words in the `allow` list. Otherwise latest `minor` release is breaking the builds.

It's a bit disappointing that their minor releases break it for us: https://github.com/wooorm/alex

### DIFF
--- a/.alexrc
+++ b/.alexrc
@@ -1,9 +1,12 @@
 {
   "allow": [
+    "children",
     "destroy",
+    "destroys",
     "color",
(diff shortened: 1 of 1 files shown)


wooorm:
@fahad19 Yeah sorry about that, I didn’t think it would affect to many words! 😬

fahad19:
@wooorm: it's alright! :D thanks for creating `alex`!

PR #224, opened 2017-06-15
Minor fix for docs in frint-react

yurist38:
Small fix for documentation.


fahad19:
thanks @yurist38! latest docs now live at: https://frint.js.org/docs/packages/frint-react/#region-and-data

PR #233, opened 2017-07-20
frint-router-react: unit tests for Link component

viacheslaff:


packages/frint-router-react/src/Link.js
```diff
@@ -11,29 +13,49 @@ export default class Link extends React.Component {
...
+  componentWillReceiveProps(nextProps) {
+    this.considerSubscribingToRouter(nextProps);
   }
 
   componentWillUnmount() {
+    this.unsubscribeFromRouter();
+  }
+
+  considerSubscribingToRouter(nextProps) {
+    if (typeof nextProps.activeClassName === 'string') {
```
fahad19: consider returning early, to avoid nesting:

```js
if (typeof nextProps.activeClassName !== 'string') {
  return;
}

// ... rest of the code
```

packages/frint-router-react/src/Link.spec.js
```diff
@@ -0,0 +1,264 @@
...
+      </Link>,
+      { context: createContextWithRouter(router) }
+    );
+
+    router.push('/');
+    expect(wrapper.hasClass('anchor')).to.be.true; // eslint-disable-line no-unused-expressions
+    expect(wrapper.hasClass('anchor-active')).to.be.false; // eslint-disable-line no-unused-expressions
+
+    router.push('/about');
+    expect(wrapper.hasClass('anchor')).to.be.true; // eslint-disable-line no-unused-expressions
```
fahad19: since too many similar inline eslint disable comments, consider moving it to the top of the file.

packages/frint-router-react/src/Link.js
```diff
@@ -1,7 +1,9 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
-export default class Link extends React.Component {
+// TODO check package(s).json
```
fahad19: if already taken care of, remove it.

packages/frint-router-react/src/Link.js
```diff
@@ -1,7 +1,9 @@
...
+
+class Link extends React.Component {
```
fahad19: one handy way of exporting the class directly:

```js
export default class Link extends React.Component {
  static propTypes = {
    foo: PropTypes.bool
  };

  render() {
    // JSX
  }
}
```

fahad19:
congrats on your first PR to FrintJS! 👏

packages/frint-router-react/src/Link.spec.js
viacheslaff: good point

packages/frint-router-react/src/Link.js
viacheslaff: removed

packages/frint-router-react/src/Link.spec.js
```diff
@@ -0,0 +1,264 @@
...
+    }
+  };
+}
+
+describe('frint-route-react › Link', () => {
+  before(function () {
+    resetDOM();
+  });
+
+  it('renders anchor with href, className and children when no type prop is passed', () => {
```
fahad19: unless needed, please consider using `function () { }` everywhere, just like how it's done in `before`.

Issue #184, opened 2017-05-05
Introduce `app.getName()`


fahad19:
App names are very frequently used, and an integral part of inter-app communication.

Instead of doing:

```js
app.getOption('name');
```

It's worth giving it a dedicated method:

```js
app.getName();
```

Since `name` in `options` is a required key.

PR #236, opened 2017-07-22
frint: new `app.getName` method

gtzio:
instead of getOption('name') use the method getName.

issue #184

### DIFF
--- a/packages/frint/README.md
+++ b/packages/frint/README.md
@@ -236,7 +236,7 @@ const RootApp = createApp({
     {
       name: 'theNameOfTheApp',
       useFactory: function ({ app }) {
-        return app.getOption('name');
+        return app.app.getName();
       },
       deps: ['app'],
       cascade: true,
@@ -303,6 +303,8 @@ The `App` instance
 #### Example
 
 `app.getOption('name')` would give you `MyAppName` string.
+
+or better `app.getName()` Since name in options is a required key.
 
 ### app.getRootApp
 
--- a/packages/frint/README.md
+++ b/packages/frint/README.md
@@ -304,7 +304,13 @@ The `App` instance
 
 `app.getOption('name')` would give you `MyAppName` string.
 
-or better `app.getName()` Since name in options is a required key.
+### app.getName
+
+> app.getName()
+
+#### Returns
+
+`String`: The App's name.
 
 ### app.getRootApp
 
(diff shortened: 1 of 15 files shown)


fahad19:
thanks for the PR!

need one more thing to do in this PR for it to pass. you can see the error message here: https://travis-ci.org/Travix-International/frint/builds/256392327#L1760

basically, we need to implement the new `getName()` method in `App`, in this file: https://github.com/Travix-International/frint/blob/master/packages/frint/src/App.js

Code that needs to be added can look something like this:

```js
App.prototype.getName = function () {
  return this.getOption('name');
};
```

gtzio:
I thought that this was implemented already :P thanks for the remark :) I will update it now

fahad19:
one thing I can recommend, always run this locally. it will let you know of any issues before it hits Travis-CI:

```js
$ npm install
$ npm run bootstrap

$ npm run lint
$ npm test
```

gtzio:
thanks a lot!!!  Definitely  I need them :)

packages/frint/README.md
```diff
@@ -304,6 +304,8 @@ The `App` instance
 
 `app.getOption('name')` would give you `MyAppName` string.
 
+or better `app.getName()` Since name in options is a required key.
```
fahad19: please also add a dedicated heading for `app.getName`. it can go after `app.getOption`: https://github.com/Travix-International/frint/blob/master/packages/frint/README.md#app-1

```md
### app.getName

> app.getName()

#### Returns

`String`: The App's name.
```

gtzio:
updated it ;)

mAiNiNfEcTiOn:
@gtziola I know it's not your fault,  but can you give a space before the `*/` in the end of this line: https://github.com/Travix-International/frint/blob/2e56d501fcc4dc7eef1caca126db27083761c53c/packages/frint/src/App.spec.js#L1

gtzio:
hahhah and I was searching why???? :) thanks Rico you save my night!!

mAiNiNfEcTiOn:
@gtzio apparently there are more lint errors on the repo. Go to sleep 😁 tomorrow is another day

gtzio:
yep I was too positive as always :P I will fix it tomorrow!

mAiNiNfEcTiOn:
@gtzio please update this branch with master to get the goodies 👍

fahad19:
💖

PR #240, opened 2017-07-26
frint-router-react › Switch component: unit tests and a little refactoring

viacheslaff:
Some changes in functionality:
- added passing of computedMath to default Route (otherwise it didn't render because there were no match)
- removed React.isValidElement check. I just couldn't come up with unit test for it. if we want to make it fool-proof, we can test it with crazy children like {' '} or () => ... or {15}, so that it would at least issue a warning that it's not a way to go.

### DIFF
--- a/packages/frint-router-react/src/Switch.spec.js
+++ b/packages/frint-router-react/src/Switch.spec.js
@@ -29,7 +29,7 @@ function createContext() {
   };
 }
 
-describe('frint-route-react › Switch', function () {
+describe('frint-router-react › Switch', function () {
   it('switches routes according to router and passes computedMatch', function () {
     const HomeComponent = () => <header>HomeComponent</header>;
     const AboutComponent = () => <article>AboutComponent</article>;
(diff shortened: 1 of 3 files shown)


packages/frint-router-react/src/Switch.spec.js
```diff
@@ -0,0 +1,287 @@
...
+      },
+    ],
+  });
+
+  return {
+    app: new App()
+  };
+}
+
+describe('frint-route-react › Switch', function () {
```
fahad19: Router? :P

fahad19:
Thanks @viacheslaff! I'll take a look later tonight.

viacheslaff:
Added checks for incorrect children types passed to `<Switch />`: particularly numbers and strings. Functions are being warned about by PropTypes. 
However if someone puts a random React element such as `<div />` it'll be treated as 'default' case and nothing past it will get rendered. I think it's okay, it's better than adding some additional runtime checks.

fahad19:
@viacheslaff: looks good! feel free to merge it to `router` branch tomorrow.

after that, kindly continue with the `guides` version of documentation, like we discussed before, where we mention about combining both `frint-router` and `frint-router-react` to have a complete working solution. one page should be enough.

bonus points, if we can also mention about Webpack's `bundle-loader` too for code splitting by route paths. but it can definitely wait. not a blocker.

PR #241, opened 2017-07-27
Guide for routing

viacheslaff:
Haven't covered following topics yet:
- Route placeholders (/route/:whatever)
- Async loading with code splitting

### DIFF
--- a/site/content/guides/routing.md
+++ b/site/content/guides/routing.md
@@ -232,24 +232,42 @@ starting with the `path`.
 ## Route params
 
 In your path you can also define params which will be passed to your component once the path is 
-matched. With params previous example can be rewritten in the following way: 
+matched. Imagine we have a photo gallery where we show a list of photo names and by clicking on the
+name you can see the name and the photo itself. Below you can find how it can be implemented by
+using route params. `PhotosPage` acts as a container with routing for two nested components: `Photos`
+and `Photo`.
 
 ```js
-const AboutAnyone = ({ match }) => <article>Some content about {match.params.party}...</article>;
-
-const AboutPage = ({ match }) => {
+const Photos = ({ match }) => {
   return (
     <div>
-      <h2>About page</h2>
+      <h1>Photos</h1>
       <ul>
-        <li><Link to={`${match.url}`}>About</Link></li>
-        <li><Link to={`${match.url}/us`}>About us</Link></li>
-        <li><Link to={`${match.url}/them`}>About them</Link></li>
+        <li><Link to={`${match.url}/1`}>Photo #1</Link></li>
+        <li><Link to={`${match.url}/2`}>Photo #2</Link></li>
+        <li><Link to={`${match.url}/3`}>Photo #3</Link></li>
       </ul>
-      <Switch>
-        <Route path={`${match.url}/:party`} component={AboutAnyone} />
-      </Switch>
     </div>
+  );
+};
+
+const Photo = ({ match }) => {
+  return (
+    <div>
+      <h1>Photo #{match.params.photoId}</h1>
+      <div>
+        <img src={`...${match.params.photoId}`} />
+      </div>
+    </div>
+  );
+}
+
+const PhotosPage = ({ match }) => {
+  return (
+    <Switch>
+      <Route path={`${match.url}/:photoId`} component={Photo} />
+      <Route component={Photos} />
+    </Switch>
   );
 };
 ```
--- a/site/content/guides/routing.md
+++ b/site/content/guides/routing.md
@@ -136,7 +136,7 @@ const RouterApp = createApp({
     {
       name: 'router',
       useFactory: function() {
-        return new HashRouterService();
+        return new BrowserRouterService();
       },
       cascade: true,
     },
--- a/site/content/guides/routing.md
+++ b/site/content/guides/routing.md
@@ -124,6 +124,7 @@ current URL. If you want to use `Link` or `Route` in your child apps you'll need
 
 ```js
 import { createApp } from 'frint';
+import BrowserRouterService from 'frint-router/BrowserRouterService';
 
 const RouterApp = createApp({
   name: 'RouterApp',
@@ -134,7 +135,9 @@ const RouterApp = createApp({
     },
     {
       name: 'router',
-      useValue: router,
+      useFactory: function() {
+        return new HashRouterService();
+      },
       cascade: true,
     },
   ],
--- a/site/content/guides/routing.md
+++ b/site/content/guides/routing.md
@@ -270,28 +270,16 @@ const PhotosPage = ({ match }) => {
 
 ## Another example
 
-For more information about usage of Frint routing you can take a look at example here: 
-https://github.com/Travix-International/frint/tree/master/examples/router 
-
-To launch it clone `frint` repository
+To learn more about usage of Frint routing you can take a look at another example from `frint` repo.
+Install `frint-cli` package, then init the `router` example and run it.
 
 ```
-git clone git@github.com:Travix-International/frint.git
-```
+$ npm install -g frint-cli
 
-Go to `frint` directory, install npm dependencies and bootstrap internal dependencies
+$ mkdir my-frint-app
+$ cd my-frint-app
 
-```
-$ cd frint
-$ npm install
-$ npm run bootstrap
-```
-
-Go to the `examples/router` directory which has the example's source code, install npm dependencies
-and start the project
-
-```
-$ cd examples/router
+$ frint init --example router
 $ npm install
 $ npm start
 ```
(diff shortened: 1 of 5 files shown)


site/content/guides/routing.md
```diff
@@ -0,0 +1,238 @@
...
+
+Because `AboutPage` is rendered using `Route` component it receives a `match` object as a prop. This
+object contains `url` which let's you define routes relative to the current.  
+
+It is important that the parent `Route` doesn't have `exact` prop so that it would match all the URLs
+starting with the `path`.
+
+## Another example
+
+For more information about usage of Frint routing you can take a look at example here:
```
fahad19: Please also add a quick command for ease of scaffolding locally:

$ frint-cli init --example router

fahad19:
This is an awesome PR! I absolutely love it how you are taking he user from just components to a functioning SPA. Brilliant! 💖

mAiNiNfEcTiOn:
I'm pretty sure I fixed that on my last PR merged here, but you have linting issues :/ Have you rebased master into your branch?

site/content/guides/routing.md
viacheslaff: I haven't dealt with cli yet. So I've added step-by-step guide with list of commands. Maybe tomorrow you could talk a bit about it.

viacheslaff:
@mAiNiNfEcTiOn, linter runs without errors on my local machine. @fahad19 told me that there were some changes to linting in master so that it does't automatically work in our branch and we should be careful when merging to `master`

fahad19:
since we have commits from 3 different contributors so far in `router` branch, we will do a rebase-merge to `master` later. we don't wanna lose others contribution info.

once this PR gets merged to `router`, we can merge `master` into `router`, and the linting issue will be gone.

site/content/guides/routing.md
```diff
@@ -0,0 +1,285 @@
...
+It is important that the parent `Route` doesn't have `exact` prop so that it would match all the URLs
+starting with the `path`.
+
+## Route params
+
+In your path you can also define params which will be passed to your component once the path is 
+matched. With params previous example can be rewritten in the following way: 
+
+```js
+const AboutAnyone = ({ match }) => <article>Some content about {match.params.party}...</article>;
```
fahad19: can we go for a more common param name? like `:user` or `:id`, that most developers can relate to right away?

site/content/guides/routing.md
```diff
@@ -0,0 +1,285 @@
...
+Go to the `examples/router` directory which has the example's source code, install npm dependencies
+and start the project
+
+```
+$ cd examples/router
+$ npm install
+$ npm start
+```
+
+It will open a new browser for you with running the example app.
```
fahad19: my English is quite limited, but may be it can be worded a bit better:

```
It will launch the example in your browser for you, after taking care of bundling.
```

site/content/guides/routing.md
```diff
@@ -0,0 +1,285 @@
...
+  );
+};
+```
+
+## Another example
+
+For more information about usage of Frint routing you can take a look at example here: 
+https://github.com/Travix-International/frint/tree/master/examples/router 
+
+To launch it clone `frint` repository
```
fahad19: what I was trying to mention before, this whole block can be shortened by showing developers how to use `frint-cli`:

```js
$ npm install -g frint-cli

$ mkdir my-frint-app
$ cd my-frint-app

$ frint init --example router
$ npm install
$ npm start
```

site/content/guides/routing.md
```diff
@@ -0,0 +1,285 @@
...
+```js
+...
+import { Route, Link, Switch } from 'frint-router-react';
+...
+const NotFoundPage = () => <div>Not found</div>;
+...
+const RootComponent = () => {
+  return (
+    <div>
+      ...
```
fahad19: better remove this line.

site/content/guides/routing.md
```diff
@@ -0,0 +1,285 @@
...
+'Home page' content. If you click the links content will change accordingly.
+
+## Default route
+
+Another useful feature of `frint-router-react` is default route which allows you to render a 
+component or an app when no other route matched. To enable this behaviour add a `Route` without a 
+`path` to your group of `Route` components and them wrap them into `Switch`.
+
+```js
+...
```
fahad19: better remove the `...`s, and have a working Component here.

site/content/guides/routing.md
```diff
@@ -0,0 +1,285 @@
...
+const RouterApp = createApp({
+  name: 'RouterApp',
+  providers: [
+    {
+      name: 'component',
+      useValue: RootComponent
+    },
+    {
+      name: 'router',
+      useValue: router,
```
fahad19: I would show how to create the instance from a `RouterService` here. preferably via `useFactory`.

site/content/guides/routing.md
viacheslaff: good point

site/content/guides/routing.md
viacheslaff: why do you think it is better inline with useFactory? it looks like more unnecessary complexity to me. also I don't understand why one would use factory here as only one instance of `router` should exist.

fahad19: the most straightforward way would be to use `useClass`.

but our RouterService classes also accept `options`, for more customizations. and `useFactory` can then take care of that use case.

if developers copy the code from here, and later they want to customize it, they would need to think less about the DI stuff we have. `useFactory` just happens to be capable of taking care of all sorts of possibilities that may arise in future for them.

site/content/guides/routing.md
viacheslaff: Thanks for explaining. Would you like to see it replacing the list of commands I wrote? I see pros and cons to both. The lists of commands are of about the same length, so I wouldn't care about this.
With cli approach you can promote usage of cli, but as a user I might be reluctant to installing some packages globally. What do you think?

fahad19: I can understand about the reluctance towards installing something globally. I think the new `npx` can be handy in this case.

But since I see you are already pointing the user to the example directory in the repo, the example's README then should tell them enough. just like kitchensink's: https://github.com/Travix-International/frint/tree/router/examples/kitchensink

We should add a README for the `router` example there too.

And in the guides here, we can keep the `frint-cli` commands. I see more tools coming around `frint-cli` in near future. It would be good to promote it wherever we can early.

viacheslaff: Ok, let's do that

site/content/guides/routing.md
viacheslaff: it's a good advantage, I agree. but you can also pass `options` to the new class instance in a very straightforward way when you create it to be later used in `useValue`:
```
const router = new BrowserRouterService(options);
```
and it doesn't add complexity by introducing factory. although I agree the complexity is pretty minimal, if we want to target as many devs as possible, I would avoid any

fahad19: if you are constructing something, and then returning the value right away, then the developer can read everything related to it in one place. which is our `router` provider.

otherwise, I have to look up and down in the file and find the relevant code to make it all understandable in my head. 

our guides should aim for an experience where developers can pick things up quickly with as less mental load needed as possible.

viacheslaff: for me the idea was to construct the app in a step by step way: first create components, then router, then app, then run it. if it was all in one place it would cause too much cognitive load (imho). so merging two sections (creating an app and a router) damages this flow a little bit.

if the developer needs to add options to router later while he has a working example in his editor, he can add them right there where it's being initialized

site/content/guides/routing.md
```diff
@@ -0,0 +1,290 @@
...
+pass them as providers to the app. Names of the providers are defined here by convention. 
+
+- `frint-react` requires that app's main component would be a provider named `component`.
+- `frint-router-react` relies on `router` provider name for `Link` and `Route` to get and set 
+current URL. If you want to use `Link` or `Route` in your child apps you'll need to cascade the 
+`router` provider to them by adding `cascade: true`   
+
+```js
+import { createApp } from 'frint';
+import BrowserRouterService from 'frint-router/BrowserRouterService';
```
fahad19: code below uses `HashRouterService`, so you may want to import that one here.

viacheslaff: :( my bad. I'd rather use BrowserRouterService. hopefully all devs are using modern broswers

fahad19:
since this PR is about docs, I would suggest also to create new markdown files, that would bring in the API docs in our site, like how it is done here for existing packages: https://github.com/Travix-International/frint/tree/master/site/content/docs/packages

and then add sidebar links here: https://github.com/Travix-International/frint/blob/master/site/partials/docsSidebar.html#L15

viacheslaff:
@fahad19, added links to READMDE docs for `frint-router` and `frint-router-react`

PR #244, opened 2017-07-28
Router

fahad19:
#### NOTE: this PR should be rebase-merged to keep multiple contributors name in the history

Will be merged by @fahad19 once approved.

---

## What's done

Introduced two new packages:

* `frint-router`
* `frint-router-react`

Documentation can be found in the diff.


packages/frint-router-react/src/Link.js
```diff
@@ -0,0 +1,118 @@
...
+    const linkProps = {
+      onClick: this.handleClick,
+      className: className || '',
+    };
+
+    if (this.state.active) {
+      linkProps.className += ` ${activeClassName}`;
+    }
+
+    if (typeof type === 'undefined') {
```
jackTheRipper: what happened if `type` is equal to `null` ?

packages/frint-router-react/src/Link.spec.js
```diff
@@ -0,0 +1,264 @@
...
+
+    expect(wrapper.children().length).to.equal(1);
+    expect(wrapper.children().get(0)).to.equal(linkContent);
+  });
+
+  it('renders button with type, className and children when type prop is passed', function () {
+    const linkContent = <span className="decorated">About</span>;
+
+    const wrapper = shallow(
+      <Link className="fancy-button" to="/about" type="button">{linkContent}</Link>,
```
jackTheRipper: `fancy-button` ❤️

packages/frint-router-react/src/Route.js
```diff
@@ -0,0 +1,113 @@
...
+export default class Route extends React.Component {
+  static contextTypes = {
+    app: PropTypes.object.isRequired
+  };
+
+  static propTypes = {
+    path: PropTypes.string,
+    exact: PropTypes.bool,
+    computedMatch: PropTypes.object,
+    component: PropTypes.oneOfType([PropTypes.node, PropTypes.func]),
```
jackTheRipper: Shouldn't we accept `PropTypes.element` as well here?

packages/frint-router-react/src/Route.js
```diff
@@ -0,0 +1,113 @@
...
+      this.setState({
+        component: nextProps.component,
+      });
+    } else if (nextProps.app && (this._appInstance === null || nextProps.app !== this.props.app)) {
+      // app
+      this._destroyRouteApp();
+
+      const RouteApp = nextProps.app;
+
+      this._appInstance = new RouteApp({
```
jackTheRipper: Couldn't we `get` the instance instead of creating a new object ?

packages/frint-router-react/src/Route.js
```diff
@@ -0,0 +1,113 @@
...
+      this._appInstance = null;
+    }
+  }
+
+  render() {
+    const ComponentToRender = this.state.component;
+    const matched = this.props.computedMatch || this.state.matched || null;
+
+    return ComponentToRender !== null && matched !== null
+      ? <ComponentToRender match={matched} />
```
jackTheRipper: just a matter of style but maybe a `if` here could be more readable

packages/frint-router-react/src/Switch.js
```diff
@@ -0,0 +1,71 @@
...
+      this._routerSubscription.unsubscribe();
+    }
+  }
+
+  render() {
+    let child = null;
+
+    React.Children.forEach(this.props.children, (element) => {
+      if (child !== null) {
+        return;
```
jackTheRipper: Shouldn't we explicitly return `null` here?

packages/frint-router/src/matchFromHistory.js
```diff
@@ -0,0 +1,40 @@
+import _ from 'lodash';
+import pathToRegexp from 'path-to-regexp';
+
+export default function matchFromHistory(pattern, history, options = {}) {
+  if (!pattern) {
+    return {
+      url: history.location.pathname,
```
jackTheRipper: Wouldn't it be better to check if `history.location` exists in the `if`?

packages/frint-router/src/matchFromHistory.js
```diff
@@ -0,0 +1,40 @@
...
+  }
+
+  const keys = [];
+  const re = pathToRegexp(pattern, keys, {
+    end: options.exact || false,
+  });
+
+  const matched = re.exec(history.location.pathname);
+
+  if (!matched) {
```
jackTheRipper: I would check as well `Array.isArray(matched)`

packages/frint-router-react/README.md
```diff
@@ -0,0 +1,333 @@
...
+
+import HomePage from './HomePage';
+import AboutPage from './AboutPage';
+
+export default function Root() {
+  return (
+    <div>
+      <Route path="/" component={HomePage} exact />
+      <Route path="/about" component={AboutPage} />
+    </div>
```
markvincze: Just to make sure I understand: so this piece of code won't render any actual content, it's just for registering routes to components? Or will this actually display something?

asci: it won't display until window url match the `path`
Really simplified version of this would be a function:
```js
function Route({path, component}){
 if (window.location.path === path) {
  return <component match={someDataAboutCurrentRouteLikeParams} />
 }
 return null;
}
```

fahad19: it will render, once the current location in the browser matches to any of the paths. if there is a match, then the assigned component (`HomePage` or `AboutPage`) will render accodingly.

packages/frint-router-react/src/Route.js
fahad19: we can't `get` because the instance does not live anywhere. these are App classes that are passed to `<Route app={App} />` components.

since they are not registered to the Root app, like how we do it for Apps targeting Regions, they can't be obtained later from somewhere else. these Apps' lifecycle are contained by the `<Route>`.

packages/frint-router-react/src/Link.js
```diff
@@ -0,0 +1,118 @@
...
+  unsubscribeFromRouter() {
+    if (this.subscription) {
+      this.subscription.unsubscribe();
+    }
+  }
+
+  handleClick = (e) => {
+    e.preventDefault();
+
+    // TODO avoid pushing route if it's current URL
```
markvincze: How big is this task, can we do it now? 😺

packages/frint-router-react/src/Link.js
```diff
@@ -0,0 +1,118 @@
...
+    if (this.state.active) {
+      linkProps.className += ` ${activeClassName}`;
+    }
+
+    if (typeof type === 'undefined') {
+      // anchor
+      linkProps.href = to;
+
+      return (
+        <a {...linkProps}>
```
markvincze: wow, this is cool, haven't seen the spread being used like this before

packages/frint-react/package.json
```diff
@@ -38,7 +38,9 @@
     "prop-types": "^15.5.10",
     "react": "^15.5.4",
     "react-dom": "^15.5.4",
-    "react-test-renderer": "^15.5.4"
+    "react-test-renderer": "^15.5.4",
+    "frint": "^2.0.1",
+    "frint-test-utils": "^2.0.1"
```
viacheslaff: `frint` and `frint-test-utils` are duplicates. one pair should be removed

packages/frint-react/package.json
```diff
@@ -38,7 +38,9 @@
     "prop-types": "^15.5.10",
     "react": "^15.5.4",
     "react-dom": "^15.5.4",
-    "react-test-renderer": "^15.5.4"
+    "react-test-renderer": "^15.5.4",
```
viacheslaff: because `enzyme`, `prop-types`, `react`, `react-dom` , `react-test-renderer` are now in root `package.json`, they can be removed this `package.json`

packages/frint-router/README.md
```diff
@@ -0,0 +1,239 @@
...
+> getMatch$(pattern, options)
+
+Keeps matching pattern against history as it keeps changing over time.
+
+#### Arguments
+
+1. `pattern` (`String`): Pattern to match against
+1. `options` (`Object`)
+  * `options.exact` (`Boolean`): Matches pattern exactly, defaults to `false`.
+  * `options.cache` (`Boolean`): Cache matches (TODO)
```
markvincze: By `TODO` you mean this is not implemented yet? Shouldn't we just remove this from the readme?

fahad19: a minimal amount of caching should be supported. repeated route matching can be costly on component re-renders. I will take care of it.

packages/frint-router-react/src/Link.js
viacheslaff: `type` prop is meant to contain `type` property for a `button` element. if somebody passes a `type` as a prop, then a `<button>` gets rendered, otherwise an `<a>`. I'm not even sure what should happen when somebody passes `null` there and what the developer expects in that case.

site/content/guides/routing.md
```diff
@@ -0,0 +1,283 @@
...
+`Link` you can choose whether it should be an exact match. You can also pass a Frint app as an `app`
+ prop instead of `component`.
+
+### Import the preferred router
+
+There are three router services available:
+
+- `BrowserRouterService`: uses modern HTML5 History API
+- `HashRouterService`: for legacy browsers
+- `MemoryRouterService`: useful for tests
```
markvincze: Which one should an average app developer use?  
Isn't there a way to automatically figure out which one is the best choice with feature detection or something?

packages/frint-router-react/README.md
markvincze: Ah ok, I think I understand, so the difference between this and using `<Switch>` is that this would simply render all the components which match, not just the first match, right?

site/content/guides/routing.md
fahad19: it's a bit difficult to decide which one should be used by default by anyone. if your application is backed by a server, then `BrowserRouterService` makes sense.

`BrowserRouterService` would keep URLs in your browser like `/about/page/here`.

But `HashRouterService` can work anywhere, as long as the target page is loaded, and URLs would look like this in the browser `/index.html#/about/page/here`.

And `MemoryRouterService` has nothing to do with the browser's API. it just imitates the behaviour of the other services and very useful in unit testing your app.

We want the developers to pick a service of their own needs when constructing the App, otherwise we end up having too much code in our bundles, that's never executed.

The developer needs to make a conscious decision about it. If you want feature detection, you have to do it yourself when returning the instance of the right service when your `router` provider is computed.

packages/frint-router-react/src/Route.js
viacheslaff: here we expect a definition of component, a class, not an instance (`Component`, not `<Component />`). that's why `PropTypes.element` doesn't fit here. so it's a function, because JavaScript :). also there can be stateless components, those are definitely functions.
actually `PropTypes.node` might not be needed. I am going to check.

packages/frint-router-react/README.md
fahad19: exactly.

`<Switch>` would always render the first matching child, and stop there. If nothing else matches, then the last child (if it happens to have no `path` prop) will be rendered.

Much like `switch` in the language, with `default` as the last case.

packages/frint-router-react/src/Link.js
fahad19: @viacheslaff: could you kindly take a look at this?

packages/frint-router-react/src/Switch.js
viacheslaff: I don't think so. Here it's works as a `continue` statement for `.forEach` loop.

packages/frint-router-react/src/Route.js
viacheslaff: both look fine to me. I wouldn't change it unless many people find it unreadable.

discosultan: I also think that both provide the same level of readability, therefore would keep it the way it is.

jackTheRipper: Yeah that's why I said a matter of style xD

packages/frint-router-react/src/Switch.js
jackTheRipper: oh yeah I didn't see it sorry

packages/frint-router-react/src/Route.js
```diff
@@ -0,0 +1,113 @@
...
+export default class Route extends React.Component {
+  static contextTypes = {
+    app: PropTypes.object.isRequired
+  };
+
+  static propTypes = {
+    path: PropTypes.string,
+    exact: PropTypes.bool,
+    computedMatch: PropTypes.object,
+    component: PropTypes.func,
```
viacheslaff: @jackTheRipper removed the `PropTypes.node`, because `PropTypes.func` is sufficient

packages/frint-router-react/src/Link.js
viacheslaff: I've added this TODO, it's the `history` library that issues a warning whenever the same route is pushed. there's a bit of discussion in here about whether it should warn or stay silent: https://github.com/ReactTraining/react-router/issues/4467. but I guess `<Link>` shouldn't push the url if it's `path` exactly matches current url

packages/frint-router/src/matchFromHistory.js
viacheslaff: I think `history.location` will always exist. At least it was always the case in unit tests with `MemoryRouterService` even before pushing any URLs in it. The idea here is to return any match if there's no pattern to match. @fahad19, am I correct?

site/content/guides/routing.md
jvantillo: Might make sense to copy this over into the docs ;)

fahad19: @jvantillo: good idea. I have copied the text, and will send a PR about how to choose the right router service in your app soon. but really want to release it first :D 

also plan to revamp the homepage with more content to deliver a better message to developers. a brainstorm meeting invitation is coming soon.

packages/frint-router/src/matchFromHistory.js
fahad19: yep.

PR #247, opened 2017-07-31
Router: caching

fahad19:
Implemented caching for route matches.


packages/frint-router/src/makeRouterService.spec.js
```diff
@@ -113,13 +113,21 @@ describe('frint-router › makeRouterService', function () {
 
             // second push
             null,
+
+            // third push
+            {
+              url: '/about',
+              isExact: true,
+              params: {},
+            },
```
viacheslaff: I'm missing here testing that caching actually works. I'd suggest that it should test if this match is actually the same instance (depending on enableCache option, so at least two tests).

viacheslaff:
Overall, looks good!

PR #265, opened 2017-08-16
Added Frint version number to exportable

### LINKED ISSUE #264: Export version number from package
Something like this:

```js
const Frint = require('frint');
console.log(Frint.version); // 2.2.0
```

That way, we can know while debugging in the browser what version we are dealing with quickly.

(I currently have no idea how this can be done)

reaktivo:
Fixes #264

### DIFF
--- a/packages/frint/src/index.js
+++ b/packages/frint/src/index.js
@@ -1,9 +1,11 @@
 import App from './App';
 import createApp from './createApp';
+import { version } from '../package';
 
 const Frint = {
   App,
   createApp,
+  version,
 };
 
 export default Frint;
--- a/packages/frint/src/index.spec.js
+++ b/packages/frint/src/index.spec.js
@@ -1,0 +1,16 @@
+/* eslint-disable import/no-extraneous-dependencies, func-names */
+/* global describe, it */
+import { expect } from 'chai';
+
+import Frint from '.';
+
+describe('frint', function () {
+  it('is an object', function () {
+    expect(Frint).to.be.an('object');
+  });
+
+  it('has a version', function () {
+    expect(Frint.version).to.be.a('string');
+    expect(Frint.version.split('.')).to.have.lengthOf(3);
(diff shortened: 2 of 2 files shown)


fahad19:
That was super quick! Thanks @reaktivo!

I have a few questions about it.

### From CommonJS user perspective:

* When we use it like a regular npm package (or via Webpack/Browserify), does it mean it will load the full object found in `package.json` while we needed only `version`?
* Can we somehow remove the extra stuff (package.json content) in our final bundle if we import from `frint` like this?

### From CDN user perspective:

* How can the version number be shipped in the `frint.min.js` file along with the release/publish to npm, which is then made available for consumption via unpkg.com?

So that the user can check version like this:

```html
<script src="https://unpkg.com/frint@latest/dist/frint.min.js"></script>
<script>
  console.log(window.Frint.version);
</script>
```

reaktivo:
You are right, unless we used rollup and some other plugins this will pull the whole `package.json`, I just had an idea and will push an update.

reaktivo:
@fahad19 What do you think about:

```
var plugins = [
  new webpack.DefinePlugin({
    FRINT_VERSION: JSON.stringify(require('./package').version)
  }),
];
```

This would of course make it unavailable as a standard npm package, but that might not be critical.

fahad19:
@reaktivo: that would work. but for making it work for dists, we need to be able to know the version before publishing to npm. that's where I got stuck.

fahad19:
I looked into React's repo, and I noticed they have a physical file that is updated with the version number: https://github.com/facebook/react/blob/master/src/ReactVersion.js

reaktivo:
I'll close this PR since I can't think of any simple solution that solves all of those cases accurately.

fahad19:
npm's `postversion` script can come handy, for updating the local `version.js` file: https://docs.npmjs.com/misc/scripts

Issue #149, opened 2017-03-23
frint-compat: Remove `createClass` message from deprecated warning


fahad19:
Because `createClass` doesn't exist in `frint`: https://github.com/Travix-International/frint/blob/master/packages/frint-compat/src/createService.js#L5

Same in `createFactory`.

fahad19:
Redundant after #267

PR #267, opened 2017-08-18
frint-compat: empty the package.

### LINKED ISSUE #266: Drop frint-compat features
Keep the package in the monorepo, but export an empty object for now only, and remove everything else in the package (including tests).

And release a new major version afterwards.

fahad19:
Closes #266


fahad19:
will be taken care of when we are about to release the next major semver.

discosultan:
Also closes https://github.com/Travix-International/frint/issues/149

Issue #242, opened 2017-07-28
frint-store: Support epics


fahad19:
As can be seen in action here: https://redux-observable.js.org/

---

# How to Epic in frint-store

## Constants

```js
// constants/index.js
export const PING = 'PING';
export const PONG = 'PONG';
```

## Action Creators

```js
// actions/pings.js
// nothing here so far
```

## Reducers

```js
// reducers/pings.js
import { PING, PONG } from '../constants';

const INITIAL_STATE = {
  isPinging: false,
};

export default function pings(state = INITIAL_STATE, action) {
  switch (action.type) {
    case PING:
      return {
        isPinging: true,
      };
    
    case PONG:
      return {
        isPinging: false,
      };
    
    default:
      return state;
  }
}
```

and root reducer:

```js
// reducers/index.js
import { combineReducers } from 'frint-store';
import pings from './pings';

export combineReducers({
  pings: pings,
});
```

## Epics

```js
// epics/pings.js
import { PING, PONG } from '../constants';

// - take an incoming PING
// - wait for 1 second
// - send a PONG
export default function pingEpic$(action$, store) {
  return action$ // this can be customized to fit our needs even more, right now very raw with plain RxJS
    .filter(action => action.type === PING)
    .delay(1000)
    .map(action => ({ type: PONG }))
}
```

and root epic:

```js
// epics/index.js
import { combineEpics } from 'frint-store';

import pings$ from './pings';

export default combineEpics(pings$, someOtherEpic$, ...);
```

## Store in App

```js
// app/index.js
import { createApp } from 'frint';
import { createStore } from 'frint-store';

import rootReducer from '../reducers';
import rootEpic from '../epics';

export default createApp({
  name: 'MyApp',
  
  providers: [
    {
      name: 'store',
      useFactory: function () {
        const Store = createStore({
          reducer: rootReducer,
          epic: rootEpic,
        });
        
        return new Store();
      }
    }
  ]
});
```

## Dispatching

Same like before:

```js
app.get('store').dispatch({ type: 'PING' });
```

## Summary

Fully backwards compatible. And one key thing to understand is, Epics are not a direct replacement of Action Creators in any way. In fact, they can even work together better.

And not to mention, doing things with Epics can lead to better handling of cancellation of async operations.

fahad19:
Closed via #271

PR #271, opened 2017-08-22
frint-store: Epics

fahad19:
## What's done

* Implemented #242 (Epics in `frint-store`)

## Todo

* [x] Example App (coming from @nunorfpt soon)

### DIFF
--- a/packages/frint-store/src/Store.js
+++ b/packages/frint-store/src/Store.js
@@ -8,6 +8,7 @@ function Store(options = {}) {
     thunkArgument: null,
     appendAction: false,
     reducer: state => state,
+    epic: null,
     enableLogger: true,
     console: console,
     ...options,
@@ -98,6 +99,10 @@ Store.prototype.dispatch = function dispatch(action) {
   )
     ? { ...this.options.appendAction, ...action }
     : action;
+
+  if (this.options.epic) {
+    // @TODO: handle epic
+  }
 
   return this.internalState$.next(payload);
 };
--- a/packages/frint-store/src/Store.js
+++ b/packages/frint-store/src/Store.js
@@ -1,6 +1,13 @@
 /* eslint-disable no-console */
 import _ from 'lodash';
-import { BehaviorSubject } from 'rxjs';
+import { Observable, Subject, BehaviorSubject } from 'rxjs';
+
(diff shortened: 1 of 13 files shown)


packages/frint-store/src/Store.js
```diff
@@ -68,12 +76,30 @@ function Store(options = {}) {
...
+  this._epicSubscription = null;
+
+  if (this.options.epic) {
+    this._input$ = new Subject();
+    this._action$ = new ActionsObservable(this._input$);
+    this._epic$ = new Subject();
+
+    this._epicSubscription = this._epic$
+      .map(epic => epic(this._action$, this, this.options.thunkArgument))
+      .switchMap(output$ => output$)
```
jackTheRipper: Why do you use `switchMap` here?

packages/frint-store/src/createStore.spec.js
```diff
@@ -442,4 +443,80 @@ describe('frint-store › createStore', function () {
...
+    });
+
+    const store = new Store();
+
+    expect(store.getState().ping.isPinging).to.equal(false);
+
+    store.getState$()
+      .take(3)
+      .scan(
+        function (acc, curr) {
```
jackTheRipper: Maybe you could use an arrow function here

packages/frint-store/src/Store.js
fahad19: because the `.map()` above is returning an ActionObservable, and further chaining should be done once that one starts emitting values.

Instead of writing it as `.map().switchMap()`, we can actually write it as `.switchMap()` directly even.

jackTheRipper: If we don't want to cancel the action then `mergeMap` should be enough, don't you think so?

packages/frint-store/src/createStore.spec.js
fahad19: unless we have single-line functions, or have functions that need access to its parent scope, we avoid using arrow functions everywhere in the Frint repo.

packages/frint-store/src/Store.js
fahad19: (discussed privately with @jackTheRipper to stick to `.switchMap` for now)

PR #273, opened 2017-08-22
frint-store: example and unit tests

nunorfpt:
# What was done

- New unit tests for combineEpics implementation
- Added epic async functionality in kitchensink example application

### DIFF
--- a/examples/kitchensink/.gitignore
+++ b/examples/kitchensink/.gitignore
@@ -1,1 +1,2 @@
 build
+node_modules
--- a/examples/kitchensink/.gitignore
+++ b/examples/kitchensink/.gitignore
@@ -1,2 +1,1 @@
 build
-node_modules
--- a/examples/kitchensink/app-color/epics/color.js
+++ b/examples/kitchensink/app-color/epics/color.js
@@ -1,0 +1,11 @@
+import {
+  CHANGE_COLOR_ASYNC
+} from '../constants';
+import { changeColor } from '../actions/color';
+
+export default function colorEpic$(action$) {
+  return action$
+  .filter(action => action.type === CHANGE_COLOR_ASYNC)
+  .delay(1000)
+  .map(action => changeColor(action.color));
+}
--- a/examples/kitchensink/app-color/epics/color.js
+++ b/examples/kitchensink/app-color/epics/color.js
@@ -5,7 +5,7 @@ import { changeColor } from '../actions/color';
 
 export default function colorEpic$(action$) {
   return action$
-  .filter(action => action.type === CHANGE_COLOR_ASYNC)
-  .delay(1000)
-  .map(action => changeColor(action.color));
+    .filter(action => action.type === CHANGE_COLOR_ASYNC)
+    .delay(1000)
+    .map(action => changeColor(action.color));
 }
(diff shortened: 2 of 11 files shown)


packages/frint-store/src/combineEpics.spec.js
```diff
@@ -0,0 +1,137 @@
+/* eslint-disable import/no-extraneous-dependencies, func-names */
+/* global describe, it */
+import { expect } from 'chai';
+import createStore from './createStore';
+import combineReducers from './combineReducers';
```
fahad19: would really like to see only `combineEpics` function being imported in this spec and unit tested. we don't need to bring in the whole Store functionality for this.

you can take a look at `combineReducers.spec.js` for example, how that function is unit tested.

To test, we just need to make sure, the combined root Epic can trigger all individual Epics, and if they can also stream further actions (if any). 

Can be done with RxJS alone.

Please check if this can be done like this:

```js
import { Subject } from 'rxjs';

import combineEpics from './combineEpics';
import ActionsObservable from './ActionsObservable'; // extract it to a file from Store.js

it('triggers all epics together', function (done) {
  const pingEpic = function (action$) {
    return action$
      .filter(action => action.type === 'PING')
      .map(() => ({ type: 'PONG' });
  };

  const rootEpic = combineEpics(pingEpic);

  const subject$ = new Subject();
  const actions$ = new ActionsObservable(subject);
  const result$ = epic(actions, store);
  const emittedActions = [];

  result.subscribe(emittedAction => emittedActions.push(emittedAction));

  subject.next({ type: 'PING' });

  expect(emittedActions).to.deep.equal([
    // PING
    // PONG
  ]);
});
```

fahad19:
nice job @nunorfpt!

left a comment about the test. we can go for one final review and merge after that is taken care of 👍

examples/kitchensink/.gitignore
```diff
@@ -1 +1,2 @@
 build
+node_modules
```
alexmiranda: this should already be present in the root `.gitignore` file

examples/kitchensink/app-color/epics/color.js
```diff
@@ -0,0 +1,11 @@
+import {
+  CHANGE_COLOR_ASYNC
+} from '../constants';
+import { changeColor } from '../actions/color';
+
+export default function colorEpic$(action$) {
+  return action$
+  .filter(action => action.type === CHANGE_COLOR_ASYNC)
```
fahad19: two spaces to the right.

packages/frint-store/src/combineEpics.spec.js
```diff
@@ -0,0 +1,35 @@
...
+    result$.subscribe((emittedAction) => {
+      emittedActions.push(emittedAction);
+    });
+
+    subject$.next({ type: 'PING' });
+
+    expect(emittedActions).to.deep.equal([
+      { type: 'PONG' }
+    ]);
+    done();
```
fahad19: `done()` needs to go inside the `.subscribe(fn)`. otherwise we risk finishing the test without any assertion.

packages/frint-store/src/Store.js
```diff
@@ -1,13 +1,7 @@
...
-import { Observable, Subject, BehaviorSubject } from 'rxjs';
-
-class ActionsObservable extends Observable {
-  constructor(source) {
-    super();
-    this.source = source;
-  }
-}
+import { Subject, BehaviorSubject } from 'rxjs';
+import ActionsObservable from './ActionsObservable';
```
fahad19: 👍

PR #277, opened 2017-08-28
Handlers for abstracting reactive components

fahad19:
## Before merge

Will be merged by @fahad19 after:

* [x] Docs are written
* [x] Test coverage is improved in new packages

---

## What's done

**NOTE**: Not a breaking change in any way.

This PR introduces two new packages, that enables any reactive rendering library (React, Preact or Vue) to integrate with FrintJS with ease.

New:

* `frint-component-utils`: Utils like `streamProps`, and functions for composing handlers
* `frint-component-handlers`: Handlers that other rendering libraries will implement (`frint-react` will consume it)

Updated:

* `frint-react`: Implemented via handlers internally

### What is a Handler?

Handlers follows a spec for the lifecycle of a reactive component, without tying itself to any specific library (like React or Vue). They are basically just objects with functions, which will be exposed as methods when composed together.

This is the default handler interface which other handlers are expected implement as needed:

```js
{
  // options
  app: null,
  component: null,

  // lifecycle: creation
  initialize() {},
  beforeDestroy() {},

  // data
  getInitialData() {},
  setData(key, value) {},
  setDataWithCallback(key, value, cb) {},
  getData(key) {},

  // props
  getProp(key) {},
  getProps() {},

  // lifecycle: mounting
  beforeMount() {},
  afterMount() {},

  // lifecycle: re-rendering
  beforeUpdate() {},
  shouldUpdate(nextProps, nextData) {},
  afterUpdate() {},

  // other
  getMountableComponent(app) {}
}
```

### Handler implementation

In the diff, you can see a `ReactHandler`, which is a plain object with only the methods needed to make a Handler compatible with React:

```js
{
  setData(key, value) {
    this.component.setState({
      [key]: value,
    });
  },
  setDataWithCallback(key, value, cb) {
    this.component.setState({
      [key]: value,
    }, cb);
  },
  getData(key) {
    return this.component.state[key];
  },
  getProps() {
    return this.component.props;
  },
  getProp(key) {
    return this.component.props[key];
  },
  getMountableComponent,
};
```

You can also see how multiple handlers are composed together and implemented in `Region` and `observe` Components in the diff.

## Why?

This enables other libraries to integrate with FrintJS as easily as possible. The monorepo here will always provide the handlers logic for maintaining the same behaviour, while others can start implementing these handlers with their preferred rendering library (Vue or Preact for example).

Within the monorepo, we will be consuming these handlers ourselves too to create React-specific packages.

## How does it impact existing users?

It doesn't. There is no breaking API change. `frint-react` still works the same way like it did before.

Unless you are a library developer, who wants to create new packages (not applications) for connecting Frint with other rendering libraries, this doesn't concern you much.

## Does it make FrintJS bloated?

Production bundle of `frint-react` will see an increase of ~600 bytes.

### Before this PR:

```
$ make list-dists
original 	 gzipped 	 file
--- 		 --- 		 ---
11.8K 		 2.8K 		 ./packages/frint-react/dist/frint-react.min.js
```

### After this PR:

```
$ make list-dists
original 	 gzipped 	 file
--- 		 --- 		 ---
15.5K 		 3.4K 		 ./packages/frint-react/dist/frint-react.min.js
```

### In the monorepo

`frint-component-handlers` was built to serve the needs of `frint-react`components (`observe` and `Region`).

Now `frint-router-component-handlers` can be created, that `frint-router-react` (`Route`, `Link`, etc) can implement.

### Outside of the monorepo

I have experimented with VueJS and it has been a successful one. Once this PR is merged and released, a new `frint-vue` repo will be created in this org here: https://github.com/frintjs

^-- those who feel like contributing to the community version of the org, please let me know and I will add you right away.


packages/frint-component-handlers/src/RegionHandler.js
```diff
@@ -0,0 +1,150 @@
...
+    }
+
+    return shouldUpdate;
+  },
+  afterUpdate(newProps = null) {
+    let name;
+    let uniqueKey;
+    let data;
+
+    if (newProps) {
```
alexmiranda:
```js
const { name = this.getProp('name'), uniqueKey = this.getProp('uniqueKey'), data = this.getProp('data') } = newProps || {};
```

:-)

PR #278, opened 2017-08-29
Make `isObservable` more generic

### LINKED ISSUE #274: frint-react: make `isObservable` more generic
## Currently

Currently, the `isObservable` function is written like this:

https://github.com/Travix-International/frint/blob/ade6369b38cefc61cd3c3bb57901bfcdbbf494b4/packages/frint-react/src/isObservable.js#L9-L20

## How it is a problem

While this works great for us for now, this also means anyone else who is importing RxJS like this would face a problem right away (see #260):

```js
```

Individually imported `Observable` wouldn't have `filter` or `map` operators as methods.

And then `isObservable()` wouldn't work as expected.

## What needs to be done

The function can be limited to checking only if `obj.subscribe` is a function, and the checks for operators (`map` and `filter`) should be removed.

DzmitryKukharuk:
While this works great for us for now, this also means anyone else who is importing RxJS like this import { Observable } from 'rxjs/Observable'; would face a problem right away

Closes #274

### DIFF
--- a/packages/frint-react/src/isObservable.js
+++ b/packages/frint-react/src/isObservable.js
@@ -7,12 +7,7 @@
  * @return boolean
  */
 export default function isObservable(obj) {
-  if (
-    obj &&
-    typeof obj.subscribe === 'function' &&
-    typeof obj.map === 'function' &&
-    typeof obj.filter === 'function'
-  ) {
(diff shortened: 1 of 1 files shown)


fahad19:
Thanks @DzmitryKukharuk! 💖

PR #280, opened 2017-08-29
frint-react: Declare app in context of Region

### LINKED ISSUE #256: frint-react: Region not working unless there is `window.app` variable
Can be reproduced here: https://codesandbox.io/s/5yp0wXoNX

Go to `./index.js`, and change this line from:

```js
const app = window.app = new RootApp();
```

to:

```js
const app = new RootApp();
```

and replace `window.app` references to `app`.

reaktivo:
Declaring the context types in the Region will allow to access the parent app that is being passed by the `observe` hoc.

Fixes #256

### DIFF
--- a/packages/frint-react/src/components/Region.js
+++ b/packages/frint-react/src/components/Region.js
@@ -15,7 +15,7 @@ export default class Region extends React.Component {
 
   static contextTypes = {
     app: PropTypes.object,
-  }
+  };
 
   constructor(...args) {
     super(...args);
--- a/packages/frint-react/src/components/Region.js
+++ b/packages/frint-react/src/components/Region.js
@@ -13,6 +13,10 @@ export default class Region extends React.Component {
     data: PropTypes.any,
   };
 
+  static contextTypes = {
+    app: PropTypes.object,
+  }
+
   constructor(...args) {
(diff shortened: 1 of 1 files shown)


packages/frint-react/src/components/Region.js
```diff
@@ -13,6 +13,10 @@ export default class Region extends React.Component {
     data: PropTypes.any,
   };
 
+  static contextTypes = {
+    app: PropTypes.object,
+  }
```
fahad19: `;`

fahad19:
Thanks a lot, @reaktivo! 

Have one really minor comment. Good to merge after that.

reaktivo:
@fahad19 I've added the missing semicolon. I was finding it quite complex to create a automated test for this. But had it working on one multiple-apps examples. Do you see a reason why [this](https://gist.github.com/reaktivo/6665990b186945f51328c18430875c8b) is not working?

fahad19:
Mocha loads everything in the same environment, and this has caused me a lot of confusion in tests over time.

I always try to call `resetDOM()` (loaded globally via `frint-test-utils/register`) and usually that solves it for me. Can you please try with that?

PR #281, opened 2017-08-29
Fix script errors on Windows platform

discosultan:
### PROBLEM

`lerna bootstrap` fails on Windows with

> '..' is not recognized as an internal or external command,
operable program or batch file.

for the command `../../node_modules/.bin/babel src --out-dir lib`.

### SOLUTION

Remove the relative path to executable and rely on *PATH* instead (managed by Lerna). The failing command is simplified to `babel src --out-dir lib`.

Note that while this works great with `lerna`, it does lose the ability to directly call `npm i` within that package if the dependency is missing from `package.json`. I think that behavior makes sense but it's something to keep in mind.

-----

### PROBLEM

`lerna run test` and `lerna run cover` fail on Windows with

>Warning: Could not find any test files matching pattern: './src/**/*.spec.js'.

for the command `mocha --colors --compilers js:babel-register --recursive './src/**/*.spec.js'`.

### SOLUTION

Remove single quotes. In case where multiple paths are required, escaped double quotes should be used instead. The failing command is changed to `mocha --colors --compilers js:babel-register --recursive ./src/**/*.spec.js`.

-----

### PROBLEM

`npm run dist` fails on Windows with

>UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 9): Error: Command failed: npm run dist 'DIST_MIN' is not recognized as an internal or external command, operable program or batch file.

for the command `DIST_MIN=1 webpack --config ./webpack.config.js` due to environment variable set in a Unix way.

### SOLUTION

In root package, take a dependency to `cross-env` package and change the failing command to `cross-env DIST_MIN=1 webpack --config ./webpack.config.js`.

-----

### PROBLEM

`npm run test` fails if an older version of `sinon-chai` is installed with

>Error: Cannot find module 'sinon'.

### SOLUTION

Set minimum required version of `sinon-chai` to `"^2.13.0"`. Previous versions do not support `sinon` `"^3.0.0"`.

-----

### VALIDATING THE PULL REQUEST

The following commands should succeed both in Unix and Windows shell environments:

```
npm i

npm run bootstrap
npm run transpile
npm run alex
npm run lint
npm run test
npm run cover
npm run dist

lerna run --scope frint-store transpile
lerna run --scope frint-store lint
lerna run --scope frint-store test
lerna run --scope frint-store cover
lerna run --scope frint-store dist
```


fahad19:
Awesome first PR! 💖 

I wonder why it had an impact on coverage though.

fahad19:
Have tested running the scripts. Can confirm everything working fine on mac.

fahad19:
Scoped scripts via Lerna is failing though. It it important to keep it working:

For example, this command:

```js
$ lerna run --scope frint-store cover
```

discosultan:
@fahad19 Agreed, I didn't test scoped scripts.

lerna.json
```diff
@@ -1,5 +1,5 @@
 {
-  "lerna": "2.0.0-beta.38",
+  "lerna": "2.1.2",
```
fahad19: is it needed in this PR?

discosultan: No. However, I think it's nice to get rid of the `beta` tag. Any reason to keep it?

Issue #215, opened 2017-06-01
frint-cli: Windows support


fahad19:
As noticed by @markvincze, it doesn't work well with `frint init` on Windows.

fahad19:
for `mkdir -p`, you can also consider https://www.npmjs.com/package/mkdirp

PR #283, opened 2017-08-30
frint-cli: fix init command on Windows platform

discosultan:
Closes https://github.com/Travix-International/frint/issues/215


fahad19:
conflicts.

discosultan:
@fahad19 Conflicts resolved.

packages/frint-cli/commands/init.js
```diff
@@ -37,17 +39,27 @@ module.exports = createApp({
...
-          ];
+          function streamFrintExampleToDir() {
+            request('https://codeload.github.com/Travix-International/frint/tar.gz/master')
+              .on('error', deps.console.error)
+              .pipe(tar.x({
+                filter: path => path.indexOf(`frint-master/examples/${example}/`) === 0,
+                strip: 3,
+                C: dir
+              }))
+              .on('error', deps.console.error)
```
jackTheRipper: duplicated handler

packages/frint-cli/commands/init.js
```diff
@@ -37,17 +39,27 @@ module.exports = createApp({
...
 
-          Promise.all(cmdPromises)
-            .then(() => deps.console.log('Done!'))
-            .catch(e => deps.console.error(e));
+          mkdirp(dir, function (error) {
+            if (error) {
+              deps.console.error(error);
+              return;
+            }
+            streamFrintExampleToDir();
```
jackTheRipper: shouldn't it be synchronized at some point using Promises?

packages/frint-cli/commands/init.js
discosultan: Since this is based on Node Stream API, the error handling works differently than in promises. The `pipe` operator does not forward errors, so they need to be handled at every step. ([ref](http://grokbase.com/t/gg/nodejs/12bwd4zm4x/should-stream-pipe-forward-errors))

I'm all ears for a more elegant solution :)

jackTheRipper: I could suggest [Pump](https://www.npmjs.com/package/pump) based on this [issue](https://github.com/nodejs/readable-stream/issues/283)

discosultan: Do you still have code duplication in mind or are you talking about cleaning up internal resource of the stream based on the issue you posted?

packages/frint-cli/commands/init.js
discosultan: Afaik, no synchronization is required. The code is executed in a fire and forget manner, but the runtime ensures in this case, that it's fully executed before the process is terminated. Should behave similarly to promises.

packages/frint-cli/commands/init.js
jackTheRipper: I was mostly thinking about the code duplication and having one error handler. But maybe not required here

fahad19:
@discosultan: did you manage to test it on Mac/Linux too with your changes?

Please update your branch, and then we can merge if all good 👍

discosultan:
@fahad19 Yes, the changes were tested both on Windows 10 and Ubuntu 16.

Issue #285, opened 2017-08-31
EditorConfig: remove `end_of_line` for Windows users


fahad19:
See previous discussion with @markvincze here: https://github.com/Travix-International/travix-appix-templates/pull/4#discussion_r118269044

Removing that line would help Windows users better.

mAiNiNfEcTiOn:
@fahad19 can't see any benefit on it. Doesn't harm Windows users AFAIK. And last but not the least, for .md you can do an exception (as you mentioned on the pr referenced above)

fahad19:
@discosultan, @markvincze: can you please explain the Git issue regarding this that you were mentioning before? thanks!

discosultan:
@mAiNiNfEcTiOn valid point and this PR is flawed *on Windows*! I'll try to shed some light. When dealing with this issue, we need to take into account git, editor and linter (which I missed).

---

For default installation of Git for Windows, the `core.autocrlf` option is `true`. This means when checking in to remote, eol is set to `LF`  and when checking out, `CRLF`.

Having `LF` enforced in `.editorconfig` makes editors such as VS Code pick this up and use it on edited files, so you end up with a mix of files using `LF` and `CRLF`.

Linter, by default, is ensuring `LF` as eol. If git has converted `LF` to `CRLF`, the editor becomes bloody red! ([ref](https://eslint.org/docs/rules/linebreak-style))

---

One downside to forcing `LF` on windows is editors such as VS that likes to work with `CRLF`. I think that is no longer an issue with VS 2017 if `.editorconfig` is present specifying eol.

When using Bash on Windows, it's actually nice if the project is using `LF` as eol when sharing files between Windows & Ubuntu environments.

---

I'd propose to actually revert the PR and introduce a `.gitattributes` file which overrides global git settings in order to prevent `core.autocrlf=true` behavior for Windows users. Ideally we'd want to do `core.autorlf=input`. ([ref](https://git-scm.com/book/en/v2/Customizing-Git-Git-Configuration#__code_core_autocrlf_code))

PR #286, opened 2017-08-31
Remove end of line setting from .editorconfig

discosultan:

`end_of_line` should not be specified, as it differs from platform to platform. A common approach is to let VCS decide this instead.

A potential future solution for `end_of_line = native` is discussed here:
https://github.com/editorconfig/editorconfig/issues/226

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -7,7 +7,6 @@ root = true
 charset = utf-8
 indent_style = space
 indent_size = 2
-end_of_line = lf
 insert_final_newline = true
 trim_trailing_whitespace = true
 


mAiNiNfEcTiOn:
_Trailing newlines in non-empty files are a common UNIX idiom. Benefits of trailing newlines include the ability to concatenate or append to files as well as output files to the terminal without interfering with shell prompts._

Why remove it and lose this benefits out-of-the-box?

fahad19:
`insert_final_newline` is still present. only `end_of_line` was removed to support Windows users better.

PR #292, opened 2017-09-05
Disable linebreak-style rule for linter

discosultan:

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -5,6 +5,6 @@
   "parser": "babel-eslint",
   "rules": {
     "no-underscore-dangle": 0,
-    "linebreak-style": "off"
+    "linebreak-style": 0
   }
 }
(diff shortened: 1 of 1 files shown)


.eslintrc
```diff
@@ -4,6 +4,7 @@
   ],
   "parser": "babel-eslint",
   "rules": {
-    "no-underscore-dangle": 0
+    "no-underscore-dangle": 0,
+    "linebreak-style": 0
```
mAiNiNfEcTiOn: Have you considered adding it to our [company's eslint config](https://github.com/Travix-International/eslint-config-travix/)?

discosultan: I think it makes sense to disable it by default in `eslint-config-travix` and rather opt into it on a per project basis if necessary for some reason.

discosultan: [PR for making the change in company's ESLint config](https://github.com/Travix-International/eslint-config-travix/pull/22)

PR #310, opened 2017-09-22
frint-store: Deprecate `thunkArgument`, and introduce `deps`

### LINKED ISSUE #282: frint-store: Rename `thunkArgument` to `deps`
### Async actions (thunks)

When we first created `frint-store`, we introduced thunks for handling async actions. And arguments that we expect to be made available to the thunk is passed via `thunkArgument` option.

```js
// actions/todos.js

// regular action
export function fetchedTodos(todos) {
  return {
    type: 'FETCHED_TODOS',
    todos: todos,
  };
}

// async action
export function fetchTodos() {
  return function (dispatch, getState, thunkArgument) {
    // fetch is asynchronously,
   actuallyFetchIt('/api/todos.json', function (response) {
     dispatch({ 
       type: 'FETCHED_TODOS', 
       todos: response.todos,
   });
  }
}
```

More can be seen here: https://frint.js.org/docs/packages/frint-store/#async-actions

### Epics

But since v2.3.0, we also introduced Epics (see #271).

And thunkArgument is accessed as follows:

```js
// epics/ping.js

function pingEpic(action$, store, thunkArgument) {
  return action$
    .filter(action.type === PING) // we only want PING actions here
    .delay(100); // lets wait for 100ms asynchronously
    .map(() => ({ type: PONG })); // after waiting, dispatch a PONG action
}
```

More can be seen here: https://frint.js.org/docs/packages/frint-store/#epics

## Proposal

Since thunk is not the only thing that exists to handle async actions, `thunkArgument` is not the right name to be used.

So this,

```js
const Store = createStore({
  thunkArgument: {},
});
```

can be changed to:

```js
const Store = createStore({
  deps: {},
});
```

We already use `deps` key in Apps for Providers, so it is more in line with that, and more generic without siding with either thunks or epics.

### Backwards compatibiliy

To keep things working in older versions too, I suggest we keep supporting `thunkArgument` if provided in options. But if there is any `deps` option available, use that instead. And in docs, use `deps` everywhere.

https://github.com/Travix-International/frint/blob/4ec44fab07992e8f1ffd63dcc7d27dcd1b3b06c7/packages/frint-store/src/Store.js#L6-L16

fahad19:
Closes #282 

## After merge

* [ ] Immediately release a new version, otherwise apps scaffolded via `frint init` would fail

### DIFF
--- a/packages/frint-store/src/createStore.spec.js
+++ b/packages/frint-store/src/createStore.spec.js
@@ -120,7 +120,7 @@ describe('frint-store › createStore', function () {
     const actions = [];
     const Store = createStore({
       enableLogger: false,
-      thunkArgument: { foo: 'bar' },
+      deps: { foo: 'bar' },
       initialState: {
         counter: 0,
       },
--- a/packages/frint-store/src/createStore.spec.js
+++ b/packages/frint-store/src/createStore.spec.js
@@ -116,7 +116,7 @@ describe('frint-store › createStore', function () {
     subscription.unsubscribe();
   });
 
-  it('dispatches async actions, with thunk argument', function () {
+  it('dispatches async actions, with deps argument', function () {
     const actions = [];
     const Store = createStore({
       enableLogger: false,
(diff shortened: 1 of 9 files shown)


packages/frint-store/src/createStore.spec.js
```diff
@@ -120,7 +120,7 @@ describe('frint-store › createStore', function () {
     const actions = [];
     const Store = createStore({
       enableLogger: false,
-      thunkArgument: { foo: 'bar' },
+      deps: { foo: 'bar' },
```
viacheslaff: I think for consistency it's better to rename all occurrences of `thunkArgument` in this test to `deps`

fahad19: that's what's done here. the `thunkArgument` line is in red :)

PR #316, opened 2017-09-26
Import RxJS and Lodash modules individually

### LINKED ISSUE #260: Import modules individually from RxJS
This is something that @jvantillo mentioned few months ago. I will put them with enough explanation below for further discussions.

## Current situation

Currently, we import and use Observables like this:

```js
import { Observable } from 'rxjs';

function doSomething$() {
  return Observable.of(1, 2, 3)
    .filter(x => x % 2 === 1)
    .map(x => x * 2);
}
```

This means the whole RxJS library is always expected to be loaded, adding to our bundle size.

## Recommended approach

```js
import { of } from 'rxjs/observable/of';
import { map } from 'rxjs/operator/map';
import { filter } from 'rxjs/operator/filter';

function doSomething$() {
  return Observable
    ::of(1, 2, 3)
    ::filter(x => x % 2 === 1)
    ::map(x => x * 2)
}
```

The snippet above makes use of the bind operator though, which is still in `stage-0`: https://github.com/tc39/proposal-bind-operator

---

Below, the discussion is from the side of Travix:

## Problem from our side

While the recommended approach works great for single bundles, our case is a bit different. We have multiple bundles for Apps, and a separate single `vendors.js` bundle that brings in all other third-party libraries (like RxJS), and used as Webpack's `externals`.

This allows all App developers to be able to use all the features of RxJS, without having to put any RxJS modules in their own bundles.

## Possible solution to the problem

To go for a smooth migration, we can keep a list of all possible RxJS import paths, and put them as `externals` in our Webpack configuration of the bundler. And until everything is imported the recommended way, we can keep loading the full RxJS library via `vendors.js`.

Once everyone has migrated, we can go further and track the usage of those imports, and reduce the `vendors.js` size significantly.

fahad19:
Closes #260 

## Before merge

* [x] Fail the build for any trace of `'lodash'` and `'rxjs'`
* [x] Test manually with one example

## After merge

* [ ] Update v3 migration guide
* [ ] Update our internal docs
* [ ] Update examples (has to wait until v3.0 release)

### Imports

All RxJS and Lodash imports are now pointing to specific modules to reduce overall bundle size.

#### Before

```js
import _ from 'lodash';
import { Observable } from 'rxjs';

_.isPlainObject({});

const foo$ = Observable.of('foo');
const capitalizedFoo$ = foo$.map(x => x.toUpperCase());
```

#### After

```js
import isPlainObject from 'lodash/isPlainObject';
import { of } from 'rxjs/observable/of'; // yes, lowercased `observable` in the path
import { map } from 'rxjs/operator/map';

isPlainObject({});

const foo = of('foo');
const capitalizedFoo$ = foo$::map(x => x.toUpperCase());
```

### Webpack configuration

We also introduced a new `frint-config` package, that will export Webpack's `externals` equivalent object for use in your bundler, so that you can keep using RxJS and Lodash via `<script>` tags too.

For custom bundler:

```js
// webpack.config.js
const externals = require('frint-config').externals;

module.exports = {
  // ...
  externals: Object.assign({}, externals, {
    // ...custom additional externals
  }),
  // ...
};
```

### Makefile commands

Two new Makefile commands have been added to list all RxJS and Lodash usage in the codebase:

```
$ make list-usage-rxjs
$ make list-usage-lodash
```

## Stats

Tests performed against our `counter` example:

### Before:

```
original     gzipped     file
---          ---         ---
1992.6K      393.3K      ./counter_before.js
944.1K       182.3K      ./counter_before.min.js
```

### After:

```
original     gzipped     file
---          ---         ---
1486.1K      305.4K      ./counter_after.js
775.5K       150.0K      ./counter_after.min.js
```

### Difference:

* Non minified: 506.5kB (87.9kB gzipped)
* Minified: 168kB (32.3kB gzipped)


packages/frint-component-utils/src/streamProps.js
```diff
@@ -1,12 +1,16 @@
-import _ from 'lodash';
-import { Observable } from 'rxjs';
+import isPlainObject from 'lodash/isPlainObject';
+import { Observable } from 'rxjs/Observable';
+import { of as of$ } from 'rxjs/observable/of';
+import { merge as merge$ } from 'rxjs/observable/merge';
+import { concatMap as concatMap$ } from 'rxjs/operator/concatMap';
+import { scan as scan$ } from 'rxjs/operator/scan';
```
viacheslaff: `Observable` and its operators (`of`, `merge`, etc) both are imported. Should we either get rid of `Observable` here (adjusting its usage in file) or don't import the operators separately and use them as part of `Observable` in this file?

packages/frint-component-utils/src/streamProps.js
```diff
@@ -80,14 +84,14 @@ class Streamer {
...
+    this._push(of$(object));
 
     return this;
   }
 
   get$() {
-    return Observable.merge(...this._observables)
-      .scan((props, emitted) => {
+    return merge$(...this._observables)
+      ::scan$((props, emitted) => {
```
viacheslaff: Subjective observation: adding `::` and `$` makes code look weird. But I see where the need comes from: glueing things together and indicating methods returning observables.

packages/frint-component-utils/src/streamProps.js
fahad19: There's a difference to take into account:

```js
// this one comes with all the operators
import { Observable } from 'rxjs';
typeof Observable.prototype.map === 'function';

// this one does not come with any operators
typeof Observable.prototype.map === 'undefined';
```

That's why when importing any RxJS types individually, we need to import the operators individually too.

packages/frint-config/src/externals.js
```diff
@@ -0,0 +1,41 @@
...
+  'lodash/padStart': { root: ['_', 'padStart'] },
+  'lodash/tail': { root: ['_', 'tail'] },
+  'lodash/take': { root: ['_', 'take'] },
+  'lodash/takeRight': { root: ['_', 'takeRight'] },
+  'lodash/zipObject': { root: ['_', 'zipObject'] },
+  'lodash/zipWith': { root: ['_', 'zipWith'] },
+
+  // rxjs
+  'rxjs/BehaviorSubject': { root: ['Rx', 'BehaviorSubject'] },
+  'rxjs/Observable': { root: ['Rx', 'Observable'] },
```
viacheslaff: Since we're importing `Observable` anyway, does it make sense to import individual operators? It won't reduce the bundle size but adds complexity to source files (assuming `Observable` includes all operators)

packages/frint-component-utils/src/streamProps.js
fahad19: yep, initially I wanted to avoid this `$` sign. but then I realized we import modules from both RxJS and Lodash.

imagine the `merge` module which is common in both `rxjs` and `lodash`. maintaining the convention of always ending  variable/function name with `$` helps avoid any clash:

```js
import merge from 'lodash/merge';
import { merge as merge$ } from 'rxjs/observable/merge';
```

packages/frint-config/src/externals.js
fahad19: I have answered this with an example code snippet here: https://github.com/Travix-International/frint/pull/316#discussion_r141275482

viacheslaff:
Very nice initiative and good job! 👍 
I'm curious how the bundle file sizes changed? Because I think that was the main of the effort, right?

fahad19:
@viacheslaff: thanks! one of the TODOs in the PR is about manually testing it against an example app. the impact will mostly be on the `vendors.js` file. 

I will post the stats here as soon as I am done.

Makefile
```diff
@@ -94,3 +94,16 @@ site-publish-only:
 	(cd ./_site && git add .)
 	(cd ./_site && git commit -am 'update site')
 	(cd ./_site && git push git@github.com:Travix-International/frint gh-pages --force)
+
+##
+# Usage stats
+#
+define list_usage_in_source
```
viacheslaff: I think setting additional ESLint rule to disallow full imports of `rxjs` and `lodash` would be beneficial (https://eslint.org/docs/rules/no-restricted-imports)

fahad19: awesome idea! I have implemented the ESLint check in most recent commit.

I would still keep these Makefile commands though. they help us quickly find out which modules we currently need. and the output of these commands is the primary source of information for creating `frint-config`.

packages/frint-component-utils/src/streamProps.js
viacheslaff: I see. With this approach the imports and the code itself become more complicated. Also `::` bind operator has to be used everywhere and it's still in `stage-0` actually: https://babeljs.io/docs/plugins/preset-stage-0/

They also suggest following approach to importing individual operators
```
import 'rxjs/add/observable/of';
import 'rxjs/add/operator/map';

Observable.of(1,2,3).map(x => x + '!!!');
```
(found here https://github.com/ReactiveX/rxjs#es6-via-npm)

The code becomes much cleaner in my opinion, but then you're stepping on a slippery path of indirect imports, which modify other modules at the runtime and you no longer know what depends on what: https://github.com/ReactiveX/rxjs/blob/master/src/add/operator/map.ts

Reminds me of jQuery plugins 😱 

I think the approach you chose is better.

packages/frint-component-utils/src/streamProps.js
viacheslaff: yes, totally see why you did that

Makefile
viacheslaff: Nice. Of course, Makefile commands are also useful

packages/frint-component-utils/src/streamProps.js
viacheslaff: By the way, since you went for the bind operator approach, shall we forbid importing `rxjs/add/*` modules as well?

fahad19: good point. I have added that check in the most recent commit.

PR #324, opened 2017-10-02
Update eslint-config-travix to latest

discosultan:
Closes https://github.com/Travix-International/frint/pull/292


packages/frint-store/src/createStore.spec.js
```diff
@@ -18,7 +18,7 @@ describe('frint-store › createStore', function () {
     expect(Store).to.be.a('function');
   });
 
-  it('returns initial state upon subscription', function (done) {
```
discosultan: `done` callback is not required in this test, since it is essentially sync.

packages/frint-store/src/combineEpics.spec.js
```diff
@@ -23,14 +23,15 @@ describe('frint-store › combineEpics', function () {
     const result$ = rootEpic(actions$);
     const emittedActions = [];
 
-    result$.subscribe((emittedAction) => {
+    const subscription = result$.subscribe((emittedAction) => {
       emittedActions.push(emittedAction);
-    }).then(done());
```
discosultan: Fixes linting error.

viacheslaff: A callback is expected there. Shouldn't it be just `done` or `() => done`?

packages/frint-store/src/createStore.spec.js
```diff
@@ -32,8 +32,6 @@ describe('frint-store › createStore', function () {
         expect(state).to.deep.equal({
           ok: true,
         });
-
-        done();
       });
 
     subscription.unsubscribe();
```
jackTheRipper: wouldn't it be cleaner to do something like:
```
it('returns initial state upon subscription', function () {
      const Store = createStore();
      const store = new Store({
        enableLogger: false,
       initialState: {
         ok: true,
       }
     });
 
     const subscription = store.getState$()
       .subscribe(function (state) {
          expect(state).to.deep.equal({
            ok: true,
          });
         return  subscription.unsubscribe();
        });
       return subscription;
});
```

If you return an async method, then `done` is optional.

discosultan: It's currently following the same pattern as rest of the tests do. @fahad19 has plans to move from `chai` to `jest` or something similar to improve asserting reactive flows. This means that the tests will be reworked soon, so there's no value in perfecting them at this point.

Issue #220, opened 2017-06-03
frint-cli: Support `new` subcommand


fahad19:
## Current situation

We currently have `frint init` command, which initializes an example app in an empty directory.

The process is as follows:

```
$ mkdir my-directory
$ cd my-directory

$ frint init
```

We require the user to create a new directory before they can run the `init`.

## Proposed subcommand

We can additionally support a `new` subcommand, which will take care of creating the directory too.

```
$ frint new example-name [my-directory]
```

We can expect the command above to create a new `my-directory`, and then initialize the kitchensink example there.

If the directory argument is not present, then it would initialize the app in current working directory.

### Sources

If the example name as no `/` in the string, we would get it from our monorepo's examples/[example-name] directory.

But we can also provide a way to initialize examples from other repositories:

```
$ frint new "frintjs-templates/foo"
```

The above command means, it will initialize an example app from https://github.com/frintjs-templates/foo repository.

discosultan:
I'd scope this particular issue to build upon the current `frint init` command so that it only adds the directory creation and working directory specific rules.

Being able to fetch examples from sources other then the official repository should apply both for `frint init` and `frint new` commands and be handled as a separate issue blocked by this.

PR #325, opened 2017-10-02
frint-cli: Introduce "new" command

discosultan:
### About

Closes https://github.com/Travix-International/frint/issues/220

Introduces the `frint new` command which will replace `frint init` in the future. For more information, see the issue above.

```sh
frint new <directory> --example <example>
```

where:

- `<directory>` (optional) - name of the app folder to create. If not specified, will populate current working dir
- `<path>` (optional, default=`frintjs/frint/tree/master/examples`) - GitHub repository path for example apps
- `<example>` (optional, default=i.) - has two different shapes:
  1. `<name>` (default=`counter`) - name of the example to fetch from FrintJS official repository
  2. `<organization>/<repository>/tree/<branch>/**` - full path to arbitrary GitHub repository to fetch an example from

------

### Testing

Can be tested running the following script:

```sh
mkdir test1
cd test1
frint new
cd ..
frint new test2
frint new test3 --example router
frint new test4 --example frintjs/frint/tree/frint-router/examples/router
```

The script will generate four test folders in the current working dir for various `frint new` scenarios.


fahad19:
please also update the README of `frint-cli`.

and also the root README of the repo under `Quickstart`.

discosultan:
@fahad19 Before updating all the docs, is the command interface final? I mean mostly naming for params: `path` & `example` and using the structure `--<param>=<value>`.

fahad19:
if `--path` is already present in the command, do we need an additional `--example` too?

fahad19:
The `<name>` option can be renamed as `<directory>` or something more descriptive to avoid any confusion. I first thought `<name>` is for example's name :)

alexmiranda:
I think both `--name` and `--example` are confusing flags in this context.  If you're creating a new project, I suppose you want to determine the `--template` instead of the location/name of the example project it will download.  Just my opinion but I'm ok with the current flags if you've had discussions on it already.

discosultan:
> if --path is already present in the command, do we need an additional --example too?

Currently they are separated in order not to always have to specify the full path with `<example>`. If you don't specify `<path>`, it will default to the official repo. Alternative would be to make `<example>` "smarter" to figure out itself whether it is defined as a full path or only an example name.

> The <name> option can be renamed as <directory> or something more descriptive to avoid any confusion. I first thought <name> is for example's name :)

I agree only because all it currently does is create and name a directory. If it were to also transform `package.json`, for example, to name the application, then `<name>` would feel better. But this is something that can be changed in the future without compat issues. Therefore, I'll rename it to `<directory>` 😁.

discosultan:
> I think both --name and --example are confusing flags in this context. If you're creating a new project, I suppose you want to determine the --template instead of the location/name of the example project it will download. Just my opinion but I'm ok with the current flags if you've had discussions on it already.

Yeah, it's a bit confusing, because the intents of scaffolding a template vs setting up an example are sort of mixed. But previous `frint init` worked the same way, so I'm happy with keeping it the way it is.

markvincze:
I kinda like the way `dotnet new` works:

 - it always scaffolds in the current directory, but can be overwritten with `--output`
 - the name by default is the name of the directory, but can be overwritten with `--name`
 - the template to use is the only non-flag argument of the command
 - (and it has some additional flags, like `--language`, `--auth`, etc.)

So you can do:

 - `dotnet new mvc`: scaffolds an MVC app in the current directory with the default name. I like that this command looks really intuitive, and usually this is what you want.
 - `dotnet new console --output ~/myapp --name foo --language f#`: scaffolds an F# Console app in the folder `~/myapp`, but uses the name `foo`

Not saying that we should do the same, just wanted to mention as an example.

fahad19:
this is the guide from Ember CLI for example: https://ember-cli.com/user-guide/#using-ember-cli

markvincze:
Also with Yeoman, the name of the template (which I think is the most important thing) is a non-flag argument.

discosultan:
@markvincze Yeah, that's the difficult part - there are so many different CLI schemes out there. This particular impl is based on a similar scheme used by Node Express framework CLI.

packages/frint-cli/commands/new.js
```diff
@@ -0,0 +1,99 @@
...
+
+          deps.console.log('Initializing...');
+
+          mkdirp(dir, function mkdirpCallback(error) {
+            if (error) {
+              deps.console.error(error);
+              return;
+            }
+            streamFrintExampleToDir();
+          });
```
markvincze: From line 69 to here it's duplicated also in the `init` command, maybe we can extract this code?

discosultan: I currently opted not to extract any code, because AFAIK, `frint init` command will be deprecated in the future and its code removed.

README.md
```diff
@@ -17,21 +17,16 @@ Install [`frint-cli`](https://frint.js.org/docs/packages/frint-cli/):
...
-
-```
-$ mkdir my-directory && cd my-directory
-```
-
 Initialize an example app:
 
 ```
-$ frint init --example kitchensink
+$ frint new my-directory --example=kitchensink
```
fahad19: do we need the `=` in the command?

we use `yargs` in `frint-cli`, which would already parse the values for us:

```
$ frint new my-directory --example kitchensink
```

discosultan: `yargs` can parse values from both structures. Anyway, I have 0 preference towards either of them, so I'm good with changing it.

packages/frint-cli/commands/new.js
fahad19: yeah, `init` will be deprecated. in face the `deprecation` message can be included in this PR too.

README.md
fahad19: thanks! current documentation is not using `=` anywhere. so would be great if we can keep it that way everywhere.

packages/frint-cli/commands/new.js
```diff
@@ -0,0 +1,104 @@
...
+
+  $ frint new myapp --example kitchensink
+  $ frint new myapp --example frint-vue/tree/master/examples/basic
+
+You can find a list of all available official examples here:
+https://github.com/Travix-International/frint/tree/master/examples
+`.trim();
+
+const invalidExampleArgText = `
+Invalid <example> value. Must be in one of the following formats:
```
fahad19: really loving the detailed error output 💖

packages/frint-cli/commands/new.js
```diff
@@ -0,0 +1,104 @@
...
+          const exampleParts = example.split('/').filter(str => str !== '');
+
+          if (exampleParts.length > 1 && exampleParts.length < 5) {
+            deps.console.error(invalidExampleArgText);
+            return;
+          }
+
+          const isFullExamplePath = exampleParts.length > 1;
+
+          const organization = isFullExamplePath ? exampleParts[0]                 : 'Travix-International';
```
fahad19: that's a lot of spaces xD

discosultan: Yeah, I don't know ... I really dig reading this "tabular" format. But linter is not too happy about it (had to disable the multi-space rule on top of the file)

packages/frint-cli/commands/new.js
```diff
@@ -0,0 +1,104 @@
...
+      useValue: descriptionText,
+    },
+    {
+      name: 'execute',
+      useFactory: function useFactory(deps) {
+        return function execute() {
+          // The <example> param has two shapes:
+          // * <name> - example name from the official Frint GitHub repository
+          // * <organization>/<repo>/tree/<branch>/** - full GitHub path to an arbitrary example
+          const example = deps.params.example || 'counter';
```
fahad19: not really your doing, it was me who set the default example value deep inside before.

but it is good practice to define things like this at the top of the file as a constance, like:

```js
const DEFAULT_EXAMPLE = 'counter';
```

until we remove the `init` command, both `new` and `init` can share this same constant.

packages/frint-cli/commands/new.js
```diff
@@ -0,0 +1,104 @@
...
+
+          // Split by '/' and filter out empty results.
+          // <example> arg might start or end with a separator.
+          const exampleParts = example.split('/').filter(str => str !== '');
```
fahad19: is this check whitespace proof?

discosultan: Nope, but such a value is currently considered invalid format with undefined behavior. I could add additional regex validation if that helps.

packages/frint-cli/commands/new.js
```diff
@@ -0,0 +1,104 @@
...
+          if (exampleParts.length > 1 && exampleParts.length < 5) {
+            deps.console.error(invalidExampleArgText);
+            return;
+          }
+
+          const isFullExamplePath = exampleParts.length > 1;
+
+          const organization = isFullExamplePath ? exampleParts[0]                 : 'Travix-International';
+          const repository   = isFullExamplePath ? exampleParts[1]                 : 'frint';
+          const branch       = isFullExamplePath ? exampleParts[3]                 : 'master';
```
fahad19: would really like to see these default values declared as constants on top of this file, like:

```js
const DEFAULT_ORG = 'Travix-International';
const DEFAULT_REPO = 'frint';
// etc...
```

discosultan: workin on it ;)

packages/frint-cli/commands/new.js
fahad19: I would extract this as a separate function:

```js
function getDownloadUrlFromExample(example) {
  if (example.indexOf('/') === -1) {
    return `https://.../tar.gz/${DEFAULT_BRANCH}`;
  }

  return customUrl;
}
```

packages/frint-cli/commands/new.js
markvincze: 👍

discosultan:
Very good point, @jackTheRipper ! We should definitely make both tools "feel" the same. We still have `frint init`, but I'm not aware of how `appix` CLI works. We might want to review both of them and make necessary adjustments before this PR gets merged.

packages/frint-cli/commands/new.js
```diff
@@ -0,0 +1,124 @@
...
+      reject(INVALID_EXAMPLE_ARG_TEXT);
+    }
+
+    // Split by '/' and filter out empty results.
+    // <example> arg might start or end with a separator.
+    const exampleParts = example.split('/').filter(str => str !== '');
+
+    const isFullExamplePath = exampleParts.length > 1;
+
+    ctx.organization = isFullExamplePath ? exampleParts[0]                 : DEFAULT_ORGANIZATION;
```
fahad19: please extract the `ctx` generation into a separate function.

```js
function getContextOrWhatever(example) {
  if (example.indexOf('/') === -1) {
    return exampleFromDefaultRepo;
  }

  return exampleFromCustomRepo;
}
```

Doing so would also help us avoid multiple ternaries below.

packages/frint-cli/commands/new.js
```diff
@@ -0,0 +1,124 @@
...
+    {
+      name: 'execute',
+      useFactory: function useFactory(deps) {
+        return function execute() {
+          deps.console.log('Initializing...');
+          Promise.resolve(deps)
+            .then(mapDepsToContext)
+            .then(createOutputDirectory)
+            .then(streamExampleToOutputDirectory)
+            .then(() => deps.console.log('Done!'))
```
fahad19: we can have some additional info here. like asking the user to run two more commands to start their application:

```
Done!

Please run these two commands to start your application:

  $ npm install
  $ npm start
```

packages/frint-cli/commands/new.js
```diff
@@ -0,0 +1,145 @@
...
+Invalid <example> value. Must be in one of the following formats:
+
+  * <name>
+  * <organization>/<repository>/tree/<branch>/**
+`.trim();
+const COMPLETION_TEXT = `
+Done!
+
+Please run these two commands to start your application:
+
```
fahad19: depending on whether `<directory>` argument was given or not, we can also add an addition `cd` command.

Two scenarios:

1: with directory:

```
$ frint new myapp

Done!

Please run:

$ cd myapp
$ npm install
$ npm start
```

2: without directory

```
$ frint new

Done!

$ npm install
$ npm start
```

Sorry for nitpicking. Just trying to come up with as many use cases as possible :)

packages/frint-cli/commands/new.js
```diff
@@ -0,0 +1,145 @@
...
+    // * <organization>/<repo>/tree/<branch>/** - full GitHub path to an arbitrary example
+    const example = deps.params.example || DEFAULT_EXAMPLE;
+    if (!example.match(/(^(\w|-)+$)|^\/?(\w|-)+(\/(\w|-)+){3,}\/?$/)) {
+      reject(INVALID_EXAMPLE_ARG_TEXT);
+    }
+
+    const isCustomExample = example.indexOf('/') >= 0;
+    resolve(isCustomExample ? getContextForCustomRepo() : getContextForDefaultRepo());
+
+    function getContextForDefaultRepo() {
```
fahad19: if this function receives `example` as an argument, then it doesn't have to be defined inside a function. they can all be in the root:

```js
function getContextForDefaultRepo(example) {
  return '';
}

function getOutputDirectory(yargs) {
  return '';
}

function getContextForCustomRepo(example) {
  return '';
}

function mapDepsToContext(deps) {
  return '';
}
```

packages/frint-cli/commands/new.js
discosultan: > Sorry for nitpicking. Just trying to come up with as many use cases as possible :)

No worries, good feedback is good!

I was thinking about it actually or alternatively adjusting the wording to specify to run the following commands in the *app folder*.

packages/frint-cli/commands/new.js
discosultan: Sure, I used nested functions more as a way of grouping rather than for making use of closures. It's a matter of preference, so it should be ideally consistent across the board.

@fahad19 , so you suggest to keep functions flattened within the file root and keep them as pure as possible?

fahad19: yes please. purer the better :)

fahad19:
@discosultan: any update on this? :)

discosultan:
I need to compare it to `appix` to ensure interface similarity. Hopefully will get to it soon :)

discosultan:
@fahad19 I think this one good to go now.

- Included the requested changes regarding flattening functions
- Improved the completion text shown to users
- Tested `new`'ing in current dir, specified dir and from custom repo

fahad19:
@discosultan: awesome work! 🎉 

this PR is gonna be highly impactful for beginners!

do you think it is wise to wait until we move to `frintjs` org before we merge this? the download URL can then point to the correct URL then.

let me know what you think.

discosultan:
@fahad19 Sure, I don't see any reason to rush it. Let's make the switch to new org, then I will update the `init` and `new` commands and then we can merge it.

fahad19:
cool! 👍

fahad19:
@discosultan: can merge as soon as the URL is updated.

discosultan:
@fahad19: will update after lunch

discosultan:
All should be done now

PR #344, opened 2017-10-18
Update React in examples

ascariandrea:
I've faced an issue with `webpack` externals definition.
To address the issue I've slightly changed how externals are merged in example's `webpack.config.js`.

If you agree with the approach I can go on with others examples.


examples/router/vendors/webpack.config.js
```diff
@@ -1,10 +1,12 @@
...
+
 module.exports = {
   entry: {
-    vendor: __dirname + '/index.js'
+    vendor: path.resolve(__dirname, 'index.js')
   },
   devtool: 'source-map',
   output: {
-    path: __dirname + '/../build/js',
+    path: path.resolve(__dirname, '../build/js'),
```
fahad19: very cool. now Windows users won't have issues :)

alexmiranda:
I didn't run into any issues with it before, but here is the documentation on webpack externals function: https://webpack.js.org/configuration/externals/#function

It does make sense to me that we change the way externals are merged based on this PR. Thanks @ascariandrea!

examples/router/webpack.config.js
```diff
@@ -34,7 +35,8 @@ module.exports = {
     ]
   },
   plugins: [],
-  externals: Object.assign({}, externals, {
+  externals: [...externals.filter(e => typeof e === 'function'), {
```
fahad19: you are right, the previous code wouldn't have worked well, because `require('frint-config').externals` is an Array now. nice catch, @ascariandrea!

Could you please try this one out, and see if it works:

```js
const externals = require('frint-config').externals;

module.exports = {
  externals: externals // covers RxJS, Lodash, React and PropTypes
    .concat([ // now FrintJS packages
      {
        'frint': {
          'root': 'Frint',
          'commonjs': 'frint',
          'commonjs2': 'frint',
          'amd': 'frint'
        },
        'frint-store': {
          'root': 'FrintStore',
          'commonjs': 'frint-store',
          'commonjs2': 'frint-store',
          'amd': 'frint-store'
        },
        'frint-model': {
          'root': 'FrintModel',
          'commonjs': 'frint-model',
          'commonjs2': 'frint-model',
          'amd': 'frint-model'
        },
        'frint-data': {
          'root': 'FrintData',
          'commonjs': 'frint-data',
          'commonjs2': 'frint-data',
          'amd': 'frint-data'
        },
        'frint-react': {
          'root': 'FrintReact',
          'commonjs': 'frint-react',
          'commonjs2': 'frint-react',
          'amd': 'frint-react'
        },
        'frint-router': {
          'root': 'FrintRouter',
          'commonjs': 'frint-router',
          'commonjs2': 'frint-router',
          'amd': 'frint-router'
        },
        'frint-router-react': {
          'root': 'FrintRouterReact',
          'commonjs': 'frint-router-react',
          'commonjs2': 'frint-router-react',
          'amd': 'frint-router-react'
        },
      },
    ]);
};
```

examples/router/webpack.config.js
```diff
@@ -1,12 +1,13 @@
...
 
 module.exports = {
   entry: {
-    root: __dirname + '/root/index.js',
+    root: path.resolve(__dirname, 'root/index.js'),
   },
   devtool: 'source-map',
   output: {
-    path: __dirname + '/build/js',
+    path: path.resolve(__dirname, 'build/js'),
```
fahad19: because of individual imports in FrintJS packages (like `rxjs/Observable`), we ran into issues before when the libraries were coming from a separate `vendors.js` file.

A new key `libraryTarget` key set to `umd` helped us sort the issue out.

for e.g., you can see this Webpack config here: https://github.com/Travix-International/frint/blob/master/packages/frint-data/webpack.config.js#L24

examples/router/webpack.config.js
ascariandrea: Yep, the proposed way should work as well, however, the duplicated keys won't be overridden.
Do we need to override keys in some cases?

fahad19: `frint-config` was primarily built for handling the dists generation of all our packages.

if overriding is needed or becomes an issue, I wouldn't mind if examples drop `frint-config` usage completely.

then the webpack config could look like this:

```js
const webpackRxjsExternals = require('webpack-rxjs-externals');

module.exports = {
  externals: [
    // rxjs/*
    webpackRxjsExternals(),

    // lodash/*
    function (context, request, callback) {
      if (request.startsWith('lodash/')) {
        const subModule = request.split('/')[1];

        return callback(null, {
          root: ['_', subModule],
          commonjs: request,
          commonjs2: request,
          amd: request,
        });
      }

      return callback();
    },

      // full imports
    {
      'lodash': {
        root: '_',
        commonjs: 'lodash',
        commonjs2: 'lodash',
        amd: 'lodash',
      },
      'rxjs': {
        root: 'Rx',
        commonjs: 'rxjs',
        commonjs2: 'rxjs',
        amd: 'rxjs',
      },
      'react': {
        root: 'React',
        commonjs: 'react',
        commonjs2: 'react',
        amd: 'react',
      },
      'react-dom': {
        root: 'ReactDOM',
        commonjs: 'react-dom',
        commonjs2: 'react-dom',
        amd: 'react-dom',
      },
      'prop-types': {
        root: 'PropTypes',
        commonjs: 'prop-types',
        commonjs2: 'prop-types',
        amd: 'prop-types',
      },
    },

    // frint imports
    {
      'frint': {
        'root': 'Frint',
        'commonjs': 'frint',
        'commonjs2': 'frint',
        'amd': 'frint'
      },
      'frint-store': {
        'root': 'FrintStore',
        'commonjs': 'frint-store',
        'commonjs2': 'frint-store',
        'amd': 'frint-store'
      },
      'frint-model': {
        'root': 'FrintModel',
        'commonjs': 'frint-model',
        'commonjs2': 'frint-model',
        'amd': 'frint-model'
      },
      'frint-data': {
        'root': 'FrintData',
        'commonjs': 'frint-data',
        'commonjs2': 'frint-data',
        'amd': 'frint-data'
      },
      'frint-react': {
        'root': 'FrintReact',
        'commonjs': 'frint-react',
        'commonjs2': 'frint-react',
        'amd': 'frint-react'
      },
      'frint-router': {
        'root': 'FrintRouter',
        'commonjs': 'frint-router',
        'commonjs2': 'frint-router',
        'amd': 'frint-router'
      },
      'frint-router-react': {
        'root': 'FrintRouterReact',
        'commonjs': 'frint-router-react',
        'commonjs2': 'frint-router-react',
        'amd': 'frint-router-react'
      },
    }
  ],
};
```

The config does end up quite big though, but at least it won't have any surprises if the developer wants to customize further.

fahad19: if we define ALL the externals in `frint-config`, including the `frint-*` packages, we risk not generating proper dists for the packages. our CDN users would face difficulties then.

for example, the dist of `frint-react` includes `frint-component-handlers` and `frint-component-utils` in it.

ascariandrea: What about an util file responsible to merge the `externals` from `frint-config` with others used in `examples`?
Then replace each `frint-config` import in `webpack.config.js`s with the import of the util file?
I think it's more maintainable on the long run

fahad19: if we can avoid functions, and just have arrays/objects coming from `frint-config`, that would be cool.

for example:

```js
// has RxJS, Lodash, React, PropType
const thirdPartyExternals = require('frint-config').thirdPartyExternals;

// all frint-* packages
const frintExternals = require('frint-config').frintExternals;

module.exports = {
  externals: [
    ...thirdPartyExternals,
    ...frintExternals,
  ],
};
```

ascariandrea: This was the first solution I came up with, but I think it would affect other parts of the project.
I would handle this "refactor" in a separate PR after having pointed out what we need to achieve.

fahad19: glad we are on the same page here!

to make sure things work without breaking anything, we can just do the following:

1. keep `require('frint-config').externals` as it is, since other packages depend on it for dists generation
1. add a new `require('frint-config').frintExternals` that lists all frint-* packages like the webpack config above in this comments thread (with root/amd/commonjs/etc keys)
1. and use both `externals` and `frintExternals` in examples' webpack config

please let me know how that sounds.

ascariandrea: Agree!

I would keep my last implementation with `getExternals` to define in one place how externals are merged.

Could be it good for you?

fahad19: I am afraid having a shared util in this monorepo for all examples wouldn't work well, since we have `frint-cli`.

Developers scaffold certain examples in their own machine like this:

```
$ frint init --example counter
```

Since the scaffolded example app in their machine would only contain a certain example's code, it will not be able to resolve `examples/getExternals.js` as it is in this PR's diff.

ascariandrea:
I rewrote the branch history to clean it up and leaving only relevant commits.

I wasn't able to fix the `kitchensink` example: it requires a refactor for changes related to `frint-store` and `rxjs` imports.
I followed the migration [guide](https://github.com/Travix-International/frint/blob/master/site/content/docs/migration/v3.md) but without success.
I will commit my changes in a sec, so you could help understand where the error could be.

fahad19:
I think this is one of the files that needed migration: https://github.com/Travix-International/frint/blob/master/examples/kitchensink/app-color/epics/color.js#L1-L11

so basically, all the methods (`filter`, `map`) need to be imported manually, and then applied using bind operator (`::`):

```js
import { filter } from 'rxjs/operator/filter';
import { delay } from 'rxjs/operator/delay';
import { map } from 'rxjs/operator/map';

export default function colorEpic$(action$) {
  return action$
    ::filter(action => action.type === CHANGE_COLOR_ASYNC)
    ::delay(1000)
    ::map(action => changeColor(action.color));
}
```

ascariandrea:
That's what I tried to do, without success.
The return `action$` then is `undefined` even after have changed the methods also in `Root.js`.

fahad19:
@alexmiranda, @nunorfpt, @viacheslaff: can you please help debug this issue? 

(I will be back on Tuesday next week)

fahad19:
@ascariandrea: I tried upgrading kitchensink separately to reproduce the issue: https://github.com/Travix-International/frint/compare/examples-fix?expand=1

while doing it, I realized `frint-config` could do with a bit more changes, created a new PR here: https://github.com/Travix-International/frint/pull/347

Once that PR is merged and released, would be easier to test again.

fahad19:
The kitchensink example is fixed here: https://github.com/Travix-International/frint/compare/examples-fix?expand=1

The `libraryTarget: "umd"` is very important in Webpack config.

Can merge this PR once all the examples are fixed according to the [diff](https://github.com/Travix-International/frint/compare/examples-fix?expand=1).

ascariandrea:
Now all examples work fine, except for a minor issue in `kitchensink`:
the `counter` indicator in `AppColor` doesn't get updated on `incrementCounter`.

I tried to resolve, but I didn't find a solution cause my `rxjs` knowledge is so limited :/

fahad19:
excuse my lack of command line skills, but you can find usage of all direct imports from `rxjs`, by running this command :)

```
$ find ./examples -iname "*.js" | grep -v -e ".spec.js" -e "/node_modules/" | xargs cat | grep "'rxjs'" | grep -v -e "window"
```

once the direct imports in examples are converted, should work fine.

ascariandrea:
There were 2 wrong imports, now it works like a charm @fahad19 !

fahad19:
@ascariandrea: well done! I will merge soon after build passes.

many thanks for taking care of this! 💖

ascariandrea:
It was a pleasure for me :)

PR #346, opened 2017-10-19
frint-config: exports

ascariandrea:
This should resolve the issue with `externals` in examples' `webpack.config.js`.

Once this PR has been merged I will update examples accordingly.

### DIFF
--- a/packages/frint-config/src/index.js
+++ b/packages/frint-config/src/index.js
@@ -2,13 +2,13 @@ import externals, {
   frint,
   lodash,
   thirdParties,
-  webpackRxJs
+  rxJs
 } from './externals';
 
 export default {
   externals,
   frintExternals: frint,
-  lodashExtrnals: lodash,
-  thisPartiesExternals: thirdParties,
-  wepackRxJsExternals: webpackRxJs
+  lodashExternals: lodash,
+  thirdPartiesExternals: thirdParties,
+  rxJsExternals: rxJs
 };
--- a/packages/frint-config/src/index.js
+++ b/packages/frint-config/src/index.js
@@ -7,8 +7,8 @@ import externals, {
 
 export default {
   externals,
-  frint,
-  lodash,
-  thirdParties,
-  webpackRxJs
+  frintExternals: frint,
+  lodashExtrnals: lodash,
+  thisPartiesExternals: thirdParties,
+  wepackRxJsExternals: webpackRxJs
 };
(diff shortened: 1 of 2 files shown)


packages/frint-config/src/index.js
```diff
@@ -1,5 +1,14 @@
...
+import externals, {
+  frint,
+  lodash,
+  thirdParties,
+  webpackRxJs
+} from './externals';
 
 export default {
   externals,
+  frint,
```
fahad19: adding a suffix of `Externals` in all exported keys would make sense. would help identifying which are meant to be used as externals.

since the package is called `frint-config`, in future it may also export other things besides just Webpack externals.

the final export here can be like:

```js
export default {
  externals,

  // new additions:
  frintExternals: frint,
  lodashExternals: lodash,
  rxjsExternals: webpackRxJS,
  thirdPartyExternals: thirdParties,
};
```

fahad19:
thanks again for your cooperation, @ascariandrea! 🎉 

have left a comment, and can merge it soon after it is taken care of :)

I will handle a release right after merge too, so you can continue with the examples in other PR.

packages/frint-config/src/index.js
ascariandrea: I didn't consider the `frint-config` would export other members in the future.
Having it declared as `const externals = require('frint-config');` in each example it was the only way I think it would be needed.

packages/frint-config/src/index.js
```diff
@@ -1,5 +1,14 @@
...
+  lodash,
+  thirdParties,
+  webpackRxJs
+} from './externals';
 
 export default {
   externals,
+  frintExternals: frint,
+  lodashExtrnals: lodash,
```
fahad19: `lodashExternals`

packages/frint-config/src/index.js
```diff
@@ -1,5 +1,14 @@
...
+  thisPartiesExternals: thirdParties,
+  wepackRxJsExternals: webpackRxJs
```
fahad19: `rxjsExternals`

we don't need a prefix for `webpack`

packages/frint-config/src/index.js
```diff
@@ -1,5 +1,14 @@
...
+  lodash,
+  thirdParties,
+  webpackRxJs
+} from './externals';
 
 export default {
   externals,
+  frintExternals: frint,
+  lodashExtrnals: lodash,
+  thisPartiesExternals: thirdParties,
```
fahad19: `thirdPartyExternals`

ascariandrea:
Sorry man, I've mistook every single key in that export 😢

PR #356, opened 2017-10-31
frint-react: className prop in Region component

### LINKED ISSUE #337: frint-react: Support `className` prop in Region component
## Currently

Regions are embedded in Components like this:

```js
<Region name="sidebar" />
```

## Proposed feature

Allow `className` to be passed too like a regular React component:

```js
<Region
  name="sidebar"
  className="region-sidebar"
/>
```

This would allow CSS to target those elements, which may come from external source (outside of the primary application):

```css
.region-sidebar {
  display: none;
}
```

ascariandrea:
I've followed the issue #337  and I have added the `className` to `Region` component.

Closes #337

### DIFF
--- a/packages/frint-react/src/components/Region.js
+++ b/packages/frint-react/src/components/Region.js
@@ -9,6 +9,7 @@ import ReactHandler from '../handlers/ReactHandler';
 
 export default class Region extends React.Component {
   static propTypes = {
+    className: PropTypes.string,
     name: PropTypes.string.isRequired,
     uniqueKey: PropTypes.string,
     data: PropTypes.any,
@@ -56,8 +57,10 @@ export default class Region extends React.Component {
       return null;
     }
 
+    const { className } = this.props;
+
     return (
-      <div>
+      <div className={className}>
         {listForRendering.map((item) => {
           const { Component, name } = item;
 
--- a/packages/frint-react/src/components/Region.spec.js
+++ b/packages/frint-react/src/components/Region.spec.js
@@ -155,7 +155,7 @@ describe('frint-react › components › Region', function () {
           <div>
             <p id="root-text">Hello World from Root</p>
 
-            <Region name="sidebar" />
+            <Region className="sidebar" name="sidebar" />
 
             <ul className="todos">
               {todos.map((todo) => {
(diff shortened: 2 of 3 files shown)


packages/frint-component-handlers/src/RegionHandler.js
```diff
@@ -53,9 +53,10 @@ export default {
               return;
             }
 
-            const regionArgs = this.getProp('uniqueKey')
-              ? [this.getProp('name'), this.getProp('uniqueKey')]
-              : [this.getProp('name')];
+            const regionArgs = [
```
fahad19: I don't think we need to modify the Handlers for this. Just updating `frint-react` package should be enough.

packages/frint-react/src/components/Region.js
```diff
@@ -56,13 +57,15 @@ export default class Region extends React.Component {
...
+
     return (
-      <div>
+      <div className={className}>
         {listForRendering.map((item) => {
           const { Component, name } = item;
 
           return (
-            <Component key={`app-${name}`} />
+            <Component className={className} key={`app-${name}`} />
```
fahad19: we can remove the `className` prop from here, and keep it only in the parent element.

packages/frint-component-handlers/src/RegionHandler.js
ascariandrea: Ok, this was simpler than I thought.
I've removed the 2 commits related to `frint-component-handlers`

packages/frint-react/src/components/Region.js
ascariandrea: Done!

packages/frint-react/src/components/Region.spec.js
```diff
@@ -353,4 +353,53 @@ describe('frint-react › components › Region', function () {
...
+    // register apps
+    window.app.registerApp(App1, {
+      regions: ['sidebar'],
+      weight: 10,
+    });
+
+
+    // verify
+    const paragraph = document.querySelector('p'); // @TODO: enzyme can be used?
+    expect(paragraph.parentElement.className).to.equal(className);
```
fahad19: 👍

packages/frint-react/src/components/Region.spec.js
```diff
@@ -353,4 +353,53 @@ describe('frint-react › components › Region', function () {
     // rootApp should not have the instance any more
     expect(window.app.getAppInstance('App', 'todo-item', 'todo-item-1')).to.equal(null);
   });
+
+  it('should accept className and pass down to rendered component', () => {
```
fahad19: in tests, we usually avoid arrow functions unless they are single-lined, or needs to access parent scope.

can be just `function () {` here.

packages/frint-react/src/components/Region.spec.js
```diff
@@ -353,4 +353,53 @@ describe('frint-react › components › Region', function () {
...
+
```
fahad19: newline can be removed.

packages/frint-react/src/components/Region.spec.js
```diff
@@ -353,4 +353,53 @@ describe('frint-react › components › Region', function () {
...
+      window.app,
+      document.getElementById('root')
+    );
+
+    // register apps
+    window.app.registerApp(App1, {
+      regions: ['sidebar'],
+      weight: 10,
+    });
+
```
fahad19: newline can be removed

fahad19:
great stuff, @ascariandrea! many thanks for the PR!

could you kindly also update the API docs in `packages/frint-react/README.md`? we need to document the usage of `className` prop for `Region` component.

ascariandrea:
Sure @fahad19! 
I've added the `className` to available props of `Region`, do you want me to add an example of its usage?
I didn't add the example cause I think it's quite obvious.

fahad19:
@ascariandrea: it's good enough as it is! :)

just merged your PR. I will go for a new release later today or tomorrow most likely.

many thanks again! 💖

ascariandrea:
Many thanks to you guys :)

You gifted me a new tshirt https://hacktoberfest.digitalocean.com/stats/ascariandrea 🎉

PR #357, opened 2017-11-04
New organization

### LINKED ISSUE #355: Organization
Lately, the number of repositories have been on the rise in this GitHub organization, and more will come in near future: https://github.com/frintjs

Latest one being the project's website, which is powering https://frint.js.org

All active contributors have already been added to the new organization. If you don't see them, it means they are not publicly showing their membership yet.

I propose moving this repository to the new organization too, so that we can keep all project specific repos there. 

Similarities can be drawn from another homegrown project called Estafette here: https://github.com/estafette

fahad19:
Now that the project has moved to its independent organization, I followed what [Bootstrap](https://github.com/twbs/bootstrap) did when they did something similar:

* License is now accredited to both Authors + Organization ([example](https://github.com/twbs/bootstrap/blob/v4-dev/LICENSE#L4))
* Authors info in manifests point to GitHub's contributors page ([example](https://github.com/twbs/bootstrap/blob/v4-dev/package.json#L15))

Closes #355

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,9 +4,9 @@
 
 #### Pull Requests
 * `frint-react`
-  * [#356](https://github.com/Travix-International/frint/pull/356) frint-react: className prop in Region component. ([@ascariandrea](https://github.com/ascariandrea))
+  * [#356](https://github.com/frintjs/frint/pull/356) frint-react: className prop in Region component. ([@ascariandrea](https://github.com/ascariandrea))
 * Other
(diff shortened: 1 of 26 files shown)


fahad19:
Coveralls seem to have recognized the organization transfer, but the PR's status checks are not reflecting that yet. not sure what needs to be done here.

fahad19:
@mAiNiNfEcTiOn: I think Coveralls/NodeSecurity need some attention from your account. I am guessing it's because you originally added Coveralls integration at organization level before. I can't do much here :)

Issue #334, opened 2017-10-12
frint-react-server: add more functions from ReactDOM


fahad19:
## Currently

Only one function is supported:

* `renderToString`

## Support these too

* `renderToStaticMarkup`
* `renderToNodeStream`
* `renderStaticToNodeStream`

Further reading: https://reactjs.org/docs/react-dom-server.html#overview

PR #361, opened 2017-11-07
frint-data: Collections made immutable

### LINKED ISSUE #354: Proposal: Make collections immutable by default in frint-data
## Currently

In `frint-data`, models are immutable by default. If you want any properties to be changed, you can only do it via custom defined methods:

```js
import { Types, createModel } from 'frint-data';

const Todo = createModel({
  schema: {
    title: Types.string,
  },

  setTitle(newTitle) {
    this.title = newTitle;
  }
});

const todo = new Todo({ title: 'Hello World' });

// has no impact
todo.title = 'updated title';

// can only be done via methods
todo.setTitle('Hello Universe');
```

But collections on the other hand exposes methods that can mutate its models:

```js
import { createCollection } from 'frint-data';

const Todos = createCollection({
  model: Todo,
});

const todos = new Todos([{ title: 'Hello World' }]);
todos.push(new Todo({ title: 'About us' })); 

// now `todos` has two models
```

## Expection

Like models, we also expect Collections to be immutable by default, and not allow any changes unless the class permits it. This would avoid any unpredictable behaviour in the app later.

I request for comments from everyone, like what API would be ideal to make this happen.

### 1: Mutating methods allowed inside custom methods only

```js
import { createCollection } from 'frint-data';
import Todo from '../models/Todo';

const Todos = createCollection({
  model: Todo,
  
  addTodo(todo) {
    // allow access to "write/delete" methods inside custom methods only
    this.push(todo);
  },
});
```

### 2: Additional property at class level allowing mutation

```js
import { createCollection } from 'frint-data';
import Todo from '../models/Todo';

const Todos = createCollection({
  model: Todo,

  // setting this to `true` gives access to `push` and other methods
  mutable: true,
});
```

fahad19:
Closes #354 

(**Note**: Not expecting to merge any time before next week)

## What's done

Like Models, Collections are now also immutable by default.

```js
import { Types, createModel, createCollection } from 'frint-data';

const Todo = createModel({
  schema: {
    title: Types.string,
  },
});

const Todos = createCollection({
  model: Todo,

  addTodo(todo) {
    // `push` and other mutating methods are only made available INSIDE custom defined methods
    return this.push(todo);
  },
});
```

Now the instances:

```js
const todos = new Todos([]);

todos.addTodo(new Todo({ title: 'First todo' }));
console.log(todos.length); // 1

// no one can access `push` and other mutating methods publicly
typeof todos.push === 'undefined';
```


packages/frint-data/README.md
```diff
@@ -342,6 +351,27 @@ const Todos = createCollection({
...
+### Immutable collections
+
+Collections are immutable by default. If you want to use built-in methods that mutate the collection, then you have to do them by defining custom methods first:
+
+```js
+const Todos = createCollection({
+  model: Todo,
+
+  addTodo(todo) {
+    // `push` and other mutating methods are only avaialble inside custom methods
```
discosultan: Grammar nazi reporting for duty!

`avaialble` -> `available`

markvincze:
I get why immutability is beneficial, and I like the idea of making it the default.

On the other hand, I think It can often be handy to have a mutable collection, so maybe we can make it a bit more convenient to create a mutable collection with "public" `push` and `pop` functions.
For example we could introduce a flag in the options of the `createCollection` to say we want to make it mutable:

```js
const mutableCollection = createCollection({
  model: Todo,
  mutable: true
});
```

And then the collection would get the `push` and `pop` functions by default. (I mean everything in the PR is fine, this would be *in addition* to what is there already.) What do you think?

Kinda similar in F# with records, for example:

```f#
type Person = {
    FirstName : string
    LastName : string
    Age : int
}
```

All fields are immutable, but with the `mutable` keyword:

```
type Person = {
    FirstName : string
    LastName : string
    mutable Age : int
}
```

the `Age` field becomes mutable.

fahad19:
@markvincze: it's a valid feature request. we can tackle it in another PR soon, targeting both Models and Collections.

I will create an Issue (Proposal) for this soon after this PR is merged and released.

mAiNiNfEcTiOn:
I'm seeing a bunch of different concepts mixed here:

`read-only` - By hiding attributes under the rule _it must have a method to write on it_ it makes the attributes read-only in case you have none. Which mean they are `public` getters, but `private` setters.

`immutable` - According to ImmutableJS, the definition is:

> Immutable data cannot be changed once created, leading to much simpler application development, no defensive copying, and enabling advanced memoization and change detection techniques with simple logic

which for me goes agains the concept of the private _setters_, unless they are forced to generate new immutable structures. That's not simple specially with _collections_ where - as I spoke with @fahad19 before - if you use immutable structures for a large collection that implies generating a new large collection every time you change it.

----

Now, if we look into the _schema_ I see that we're defining types for each attribute of the model. Why not extending it to:

```js
{
  schema: {
    attributeName: {
      initialValue: [Function, scalar, ...], // in case is readOnly, allow you to have a fixed value
      readOnly: [true, false], // default true, forcing the need to have a private function
      type: Types.[type], // was already there, obvious use =)
    },
}
```

* initialValue - would allow us to say that a certain attribute is `readOnly` and therefore - even if you have a setter - it is not allowed to set it (good for things like creation dates or other types of fields that should not change during execution or mutation of the remaining fields.
It can be a `Function` that, when executed, returns the initial value. E.g.:
```js
(initialModelData) => {
  return new Date.now();
}```

* readOnly - defines if an attribute is read-only. That prevents the field from being changed. Even with a setter.

----

Now, beyond of the basic functionality above, this could bring some extra logic to the model regarding the setters. If a field is not `readOnly` it means you are allowed to change its value... If that's the case, you can verify if the field as a setter defined in the schema. E.g. `setTitle()`.
If not, the model can immediately provide a _generic_ setter that the only thing it does is just receiving the value and setting the value (after validating the value against the schema).

----

And last but not the least, I'd like to do a question regarding the usage _context_ of a model. Let's take the CRUD interfaces as an example... Normally there are things that are insertable on __C__reation that are not editable/updatable on __U__pdate. So, the schema changes based on it. How are we handling that? Because it seems that I have to have 2 different schemas for the different cases.

fahad19:
unless React v16 features (like #334) land any time soon, I will merge this PR later this week and release v5.

otherwise we can wait for merging multiple PRs to master, and release one single breaking semver.

PR #363, opened 2017-11-08
Adds coverage threshold check

mAiNiNfEcTiOn:
# What does this PR do:

* Adds a new shell script `check-coverage.sh` which based on the output from `istanbul-combine` it parses the _Lines_ coverage number and verifies against an environment variable `EXPECTED_LINES_COVERAGE_PERCENTAGE` (when not provided, becomes _50_).
* Adds a new `npm` task to the `package.json` named `cover:check`, which is used in `.travis.yml` in place of the previous usage of `istanbul-combine`.

Usage:

`EXPECTED_LINES_COVERAGE_PERCENTAGE=80 npm run cover:check`

or 

`npm run cover:check 80`

or

`npm run cover:check` which is the same as `npm run cover:check 50`

  - Diffs


check-coverage.sh
```diff
@@ -0,0 +1,10 @@
+LINES_COVERAGE_PERCENTAGE=`./node_modules/.bin/istanbul-combine -d coverage -p summary -r lcov -r html packages/frint*/coverage/coverage*.json | grep Lines | awk '{print $3}' | tr -d '[:space:]' | tr -d '%'`;
```
fahad19: could you kindly move this file to a generic directory, like `scripts`?

(based on how facebook/react has it in their repo for additional scripts)

mAiNiNfEcTiOn: sure ... doing it now

jvantillo:
Might as well set the minimum code coverage to 80?

mAiNiNfEcTiOn:
@jvantillo for now it's set as 97% on travis-ci:

screen shot 2017-11-08 at 15 00 18

In case you want to decrease it, it's just a matter to change the setting on [travis-ci.org](https://travis-ci.org/frintjs/frint/settings)

PR #379, opened 2017-11-22
Fix conflict between render and component props in Route component

viacheslaff:
This commit fixes following bug. Whenever `<Route />` instance had `component` prop, and then props change to having `render` prop and not having `component` prop, the `component` would still get rendered (ignoring `render` prop), because if was saved in state and `component` state has priority over `render` prop.

Example code where it was breaking (while switching between routes):

```js
<Switch>
  <Route path="/one" component={OneComponent} />
  <Route path="/two" render={() => <AnotherComponent />} />
</Switch>
```

Now it's fixed by setting `component` state to `null` whenever `render` prop is present.

### DIFF
--- a/packages/frint-router-component-handlers/src/createRouteHandler.js
+++ b/packages/frint-router-component-handlers/src/createRouteHandler.js
@@ -51,7 +51,10 @@ const RouteHandler = {
   },
 
   _calculateComponentState(nextProps, appChanged = false) {
-    if (nextProps.component) {
+    if (nextProps.render) {
+      // render
+      this.setData('component', null);
+    } else if (nextProps.component) {
       // component
       this._destroyRouteApp();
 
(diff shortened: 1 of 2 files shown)


packages/frint-router-component-handlers/src/createRouteHandler.js
```diff
@@ -51,7 +51,10 @@ const RouteHandler = {
   },
 
   _calculateComponentState(nextProps, appChanged = false) {
```
alexmiranda: maybe this method should be refactored to return earlier? i.e. check for `component` in `nextProps`, perform the logic and return. The same for the other conditions. Not a big deal though.

fahad19: 👍

PR #383, opened 2017-12-08
Ensure proper `this` context in lifecycle callbacks

viacheslaff:
When `initialize` is defined both at the time of creation and instantiation of the apps, they didn't have proper `this` context so it couldn't for example access providers with `this.get(...)`. This PR fixes the issue.

### DIFF
--- a/packages/frint/src/App.spec.js
+++ b/packages/frint/src/App.spec.js
@@ -511,4 +511,35 @@ describe('frint  › App', function () {
     const app = new Root();
     expect(app.get('foo')).to.equal('original foo [updated]');
   });
+
+  it('can access and update providers from lifecycle callback defined while instantiating', function () {
+    const Root = createApp({
+      name: 'RootApp',
+      providers: [
+        {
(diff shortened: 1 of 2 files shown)


packages/frint/src/App.spec.js
```diff
@@ -511,4 +511,35 @@ describe('frint  › App', function () {
...
+        this.container.register({
+          name: 'foo',
+          useValue: `${foo} [updatedFromCreateApp]`,
+        });
+      }
+    });
+
+    const app = new Root({
+      initialize() {
+        const foo = this.get('foo');
```
viacheslaff: This is where it used to crash with `TypeError: this.get is not a function`

fahad19:
@viacheslaff: thanks for taking care of this! 💖

fahad19:
I am curious though how it worked in our internal project. I had screenshots in the internal PR.

viacheslaff:
@fahad19 it was mostly working, but gordian app was crashing because of this issue

PR #395, opened 2017-12-28
frint-data: support Date as a type

fahad19:
## What's done

Added a new `Types.date` in the API.

## Usage

```js
import { createModel, Types } from 'frint-data';

const Todo = createModel({
  schema: {
    createdAt: Types.date,
  },
});
```

Can be instantiated with either Date objects, or a string that `Date.parse()` can understand:

```js
const todo1 = new Todo({
  createdAt: new Date(),
});

const todo2 = new Todo({
  createdAt: 'Thu Dec 28 2017 08:04:56 GMT+0000 (UTC)',
});
```

When using `model.toJS()`, it will return the value as a string according to `date.toISOString()`:

```js
const createdAt = someTodo.toJS().createdAt; // `2017-12-28T16:29:26.258Z`
```

### DIFF
--- a/packages/frint-data/src/createModel.js
+++ b/packages/frint-data/src/createModel.js
@@ -62,6 +62,14 @@ export default function createModel(options = {}) {
                 return convertToJS(v);
               }
 
+              // for supporting Date objects
+              if (
+                v &&
+                typeof v.toISOString === 'function'
+              ) {
+                return v.toISOString();
+              }
+
               return v;
             });
           }
--- a/packages/frint-data/src/createModel.js
+++ b/packages/frint-data/src/createModel.js
@@ -65,7 +65,7 @@ export default function createModel(options = {}) {
               // for supporting Date objects
               if (
                 v &&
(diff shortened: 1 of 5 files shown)


packages/frint-data/src/createModel.js
```diff
@@ -62,6 +62,14 @@ export default function createModel(options = {}) {
                 return convertToJS(v);
               }
 
+              // for supporting Date objects
+              if (
+                v &&
+                typeof v.toISOString === 'function'
```
jackTheRipper: Why don't you use `instanceof Date` here? If it's true then it implies that you have `toISOString`

fahad19: I will update it.

fahad19:
@jackTheRipper: updated it to check if value is an instance of Date.

There seems to be decent browser support for it too, but would be nice to get @mAiNiNfEcTiOn's input about `Date.toISOString()`.

PR #396, opened 2017-12-29
Fix the signature of the function getApps$ in the d.ts

markvincze:
I started experimenting with migrating one of the packages to TypeScript, and discovered some incorrect signatures in the `d.ts` typings (there might be more) in the `getApps$` function:

 - the `regionName` argument should be optional
 - the `Observable` in the result is returning an array, not just a single app

Without the fix we get the following errors:

image

and

image

### DIFF
--- a/packages/frint/index.d.ts
+++ b/packages/frint/index.d.ts
@@ -56,7 +56,7 @@ export class App {
 
     getAppOnceAvailable$(name: string, region?: string, regionKey?: string): Observable<App>;
 
-    getApps$<T extends RegisteredApp>(regionName: string): Observable<T>;
+    getApps$<T extends RegisteredApp>(regionName?: string): Observable<T[]>;
 
     getContainer(): Container;
 


fahad19:
Awesome! Thanks for the PR, @markvincze! 💖

markvincze:
@fahad19 if I wanna release this, I should merge this and afterwards locally on `master` do `make release` with `minor`?
And that'll bump the version numbers locally, so I have to then push that as a commit to `master`?

fahad19:
This PR alone can be considered a `patch`.

But currently in `master`, we have some merged PRs already with `minor`. So our next release has to be a `minor` one via `$ make release` anyways.

fahad19:
Process:

* Merge PRs first to `master`
* Then run `$ make release`
* Choose appropriate semver from list (it will prompt automatically)
* It will continue with the release to npm, updating the version numbers everywhere, do a git tag, and also push it to GitHub all from the same command.

The only manual step required afterwards, is the generating of changelogs.

PR #398, opened 2018-01-02
frint: Support listening to Child App registrations from providers

fahad19:
## What's done

From providers, it was not possible to listen to Child App registrations before, because the dependencies were resolved BEFORE defining the Subject for apps list.

The Subject has been moved above now, without causing any unintended side-effects.

## Usage

```js
const Root = createApp({
  name: 'RootApp',
  providers: [
    {
      name: '__EXEC__',
      useFactory({ app }) {
        // before, it was not possible to call `app.getApps$()` here
        app.getApps$().subscribe(x => console.log(x);
      },
      deps: ['app'],
    },
  ],
});
```

### DIFF
--- a/packages/frint/src/App.js
+++ b/packages/frint/src/App.js
@@ -60,6 +60,10 @@ function App(opts = {}) {
     throw new Error('Must provide `name` in options');
   }
 
+  // children - create Observable if root
+  this._appsCollection = [];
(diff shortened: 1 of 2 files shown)


fahad19:
The "Node Security" check is not working since last few PRs. anyone got any clue?

Issue #388, opened 2017-12-18
 Should we move to TypeScript?


fahad19:
This is an issue for collecting feedback on whether writing FrintJS source in TypeScript is a good idea or not.

Some questions that I would like everyone to answer:

* Does it add value? How?
* What are the disadvantages?
* Our manually written typings are out of sync, would it help maintenance by moving to TS?

If all positive:

* Who can volunteer to help?

markvincze:
In short: I'd absolutely support it.

Pros:
 - Makes code more reliable, the static compilation is a first safety net before unit tests are run.
 - Type definition files don't have to be maintained manually.  
(And maybe the structured comments are also better picked up by the tooling this way? I'm not sure)
 - It's easier to jump in development and discover what certain types/functions can do due to the type system holding our hands.

Cons:
 - Some TypeScript knowledge will be needed to do development on frint, this might discourage people who don't know or like it.  
(On the other hand it might make frint more interesting for TypeScript-fans :slightly_smiling_face:) 
 - Maybe the webpack/lerna/etc. config has to become more complicated?

I'd be happy to help, but can't make promises, I'm not sure how much free time I'll have :/.

asci:
Any particular way you want to write it (like C#ish way or more Flow-ish)?

We have an experience using TypeScript in our projects and what I see so far — it really matters which features will you use. Sometimes doesn't look like JS and new concepts could lead to some new issues.

Also, tooling is matter — to get most of the typescript is it better to use VSCode\Webstorm. But there still could be some problems: for example, we found an issue with tslint using async\await. 

So for me would be cool to see the way you want to use it (like a guide or so)

discosultan:
Overall I favor the idea and I think it makes sense since we already support TypeScript through type definition files.

Pros:

- It is easy to forget to update type definitions. Confusing consumers with stale definitions is worse than not having them at all. With source being in TypeScript, we lose this problem completely.
- Very nice development experience for newcomers. Editors such as VS Code can provide excellent 
autocompletion, navigation and refactoring support.
- Less bugs due to checking a large set of problems at transpile time as opposed to runtime.

Cons:

- The source code becomes a lot more verbose.
- Requires additional TypeScript tooling to be setup. No everyone is a fan of TypeScript and some external contributors may be turned off by this.

fahad19:
Refs #406

fahad19:
Yes, we are migrating.

PR #400, opened 2018-01-09
Migrate frint-cli to TypeScript

markvincze:
This is a starting point to move `frint` to TypeScript.
I started with `frint-cli`, because it's less critical, and it's self-contained. If this looks good and we agree to start moving to TS, then the core `frint` package would be a good next candidate.

Although `frint-cli` is really simple and its code itself is not really utilizing TS, the tooling is already valuable. For example these are producing errors at compile time without running any tests, while when using JS, we don't see anything.

image

image

And we get proper autocomplete too out of the box, and we don't have to maintain `d.ts` files manually:

image

The exciting part is going to be migrating `frint`, because there we'll have to make interesting decisions about how to design the TS type model.


packages/frint-cli/package.json
```diff
@@ -2,15 +2,14 @@
...
-    "frint": "./bin/frint.js"
+    "frint": "./lib/bin/frint.js"
   },
   "scripts": {
-    "lint": "cross-env ../../node_modules/.bin/eslint --color '{root,commands,bin}/**/*.js'",
-    "transpile": "echo \"Nothing to transpile\"",
-    "test": "cross-env ../../node_modules/.bin/mocha --colors --recursive ./commands/*.spec.js",
+    "lint": "cross-env ./node_modules/.bin/tslint -p .",
+    "transpile": "cross-env ./node_modules/.bin/tsc",
+    "test": "cross-env ../../node_modules/.bin/mocha --require ts-node/register --colors --recursive ./src/**/*.spec.ts",
```
markvincze: There are multiple ways to run TS unit tests.  
We could also simply `transpile` first and then run `mocha` on the transpiled JS, but this option to use `ts-node` seems to be more popular (and this way `watch` can work properly too).

package.json
```diff
@@ -3,7 +3,7 @@
   "description": "Framework for building front-end apps",
   "private": true,
   "scripts": {
-    "bootstrap": "lerna bootstrap",
+    "bootstrap": "lerna bootstrap --hoist",
```
markvincze: The `--hoist` flag is needed due to https://github.com/Microsoft/TypeScript/issues/20945

markvincze:
The CI build is failing due to linting, I'll check that when I'll have time.

fahad19:
💖💖💖

Many thanks for leading this TypeScript initiative, @markvincze!!!

(Will check this PR out in more details toorrow)

discosultan:
merge

fahad19:
Refs #388

PR #403, opened 2018-01-14
frint-di: move it into monorepo

fahad19:
## What's done

* Imported `travix-di` (which is a fork of `diyai`), into the monorepo

## Why

* The package was previously left almost unattended
* Unit tests coverage was not up to the standards of the monorepo
* It is the base of `frint` package itself as a dependency and extremely important
* This should also help us migrate to TypeScript swiftly

## Before merge

* [x] Check coverage


markvincze:
The `index.d.ts` file was not correct in `travix-di`, you can replace it with this: https://gist.github.com/markvincze/d4cbf940687f676bcf6b4e19c7a7b0f3 (and remove `createClass` if that's not there any more).

fahad19:
Thanks for the gist, @markvincze!

I am probably doing something wrong myself, but updating the typings is now failing the build for some reason. I will investigate further from office tomorrow.

fahad19:
It was failing in `$ npm run bootstrap`.

I removed the `I` prefix from typings, and it started working again.

Cannot confirm about idiomatic TypeScript myself, but there is this Issue here that may help understand more: https://github.com/Microsoft/TypeScript-Handbook/issues/121

Issue #182, opened 2017-05-05
Drop `region` usage from `frint` package


fahad19:
Region is a feature mainly found in `frint-react` package. So `frint` package itself doesn't need to have it baked in.

PR #406, opened 2018-01-20
Port the frint package to TypeScript

markvincze:
First working version of `frint` in TS.  
I might still clean things up a bit, and I wanna see if the CI build is green.

The API of `frint` has not been changed, the `kitchesink` sample is still working with the new version.
The internal implementation is also the same, I didn't make any significant changes.

There are a couple of things that don't fit nicely into TypeScript, namely the `frintAppName` property, which we put on the class itself (so that it's a property of not instances, but the class object), and the way that `createApp` always dynamically creates a new class object, that's a bit weird to represent in TS. That's what the `AppClass` interface addresses, but I don't know how much of a usual practice that is.


packages/frint/index.d.ts
```diff
@@ -1,95 +0,0 @@
-/** Declaration file generated by dts-gen */
```
markvincze: This is not needed, since `tsc` can automatically generate all the exact type definitions.

packages/frint/src/App.ts
```diff
@@ -0,0 +1,410 @@
...
+
+    if (index === -1) {
+      throw new Error(`No app found with name '${name}'.`);
+    }
+
+    const w = this._appsCollection[index];
+    const key = makeInstanceKey(region, regionKey, w.multi);
+
+    this._appsCollection[index].instances[key] = new w.appClass({
+      ...omit(w, ['AppClass', 'instances']) as AppOptions,
```
markvincze: This is also not too nice. We're dynamically cherrypicking some of the properties of the object, so `tsc` can not figure out the type of the result, so I have to cast to `AppOptions` explicitly.

markvincze:
Build fails during coverage calculation, which I haven't tested locally. I'll fix it.

packages/frint-cli/tsconfig.json
```diff
@@ -1,8 +1,8 @@
 {
```
fahad19: is it possible to have one single `tsconfig.json` in the root of the repository?

packages/frint-di/README.md
```diff
@@ -47,7 +47,7 @@ Let's start by defining a simple container first.
 When defining the providers, we can directly assign values for them via the `useValue` key.
 
 ```js
-import { createContainer, resolveContainer } from 'travix-di';
```
fahad19: thanks for fixing this! <3

packages/frint/src/App.ts
```diff
@@ -0,0 +1,410 @@
+import { Container, createContainer, Provider, resolveContainer } from 'frint-di';
+import { find, findIndex, get, omit } from 'lodash';
```
fahad19: we tend to avoid importing directly from the library to help reduce the bundle size of our custom applications.

this code below will import the whole lodash lib:

```js
import { get } from 'lodash';
```

this code below will import only the `get` function alone:

```js
import get from 'lodash/get';
```

with tree-shaking, we could eliminate this problem. but when the custom application is composed of various bundles coming in from multiple places (repos, apps), it makes things difficult with tree-shaking. since the bundler cannot see everything from a single entry point.

packages/frint/src/App.ts
```diff
@@ -0,0 +1,410 @@
...
+import { BehaviorSubject, Observable } from 'rxjs';
```
fahad19: similar to the `lodash` comment above, please import them individually:

```js
import { BehaviorSubject } from 'rxjs/BehaviorSubject';
```

Sidenote: the import paths for these basic types will change to `rxjs/create` in RxJS v6 (which is in alpha now)

packages/frint/src/App.ts
```diff
@@ -0,0 +1,410 @@
...
+import { concatMap as concatMap$, find as find$, first as first$, map as map$ } from 'rxjs/operators';
```
fahad19: to help reduce bundle size, please consider importing the modules individually:

```js
import { concatMap } from 'rxjs/operators/concatMap';
```

---

It was indeed me that started the pattern of importing operators with a name ending `$`, but I really see no value in doing so now. It was mostly done to avoid conflicts with lodash functions, which I don't remember facing anywhere.

packages/frint/src/App.ts
```diff
@@ -0,0 +1,410 @@
...
+
+    return parents.pop();
+  }
+
+  public getParentApp() {
+    return this.options[this.options.providerNames.parentApp] as App || null;
+  }
+
+  public getParentApps() {
+    function findParents(app: App, parents: App[] = []): App[] {
```
fahad19: if there is no usage of `this` inside this function, please feel free to move this function out of this method. can stay in the same file as a private function that is not exported.

packages/frint/src/App.ts
fahad19: you are right, this part is not very well thought out.

we can combine efforts when we tackle #182. the API for internal registry of Child Apps can be done the right way then.

packages/frint/src/createApp.ts
```diff
@@ -1,6 +1,6 @@
-import merge from 'lodash/merge';
+import { merge } from 'lodash';
```
fahad19: we have an ESLint rule for avoiding importing for `lodash` and `rxjs` directly here: https://github.com/frintjs/frint/blob/master/.eslintrc#L10

may be we can do something similar with TSLint too?

packages/frint/src/App.ts
```diff
@@ -0,0 +1,410 @@
...
+export interface AppOptions {
+  name?: string;
+  parentApp?: App;
+  providers?: FrintProvider[];
+  providerNames?: ProviderNames;
+  initialize?: () => void;
+  beforeDestroy?: () => void;
+}
+
+export interface AppClass {
```
fahad19: it feels so so good to read these interfaces :')

packages/frint-cli/tsconfig.json
markvincze: I looked into it: there is a way to have a base config, and then extend it, but it's not working to me properly.  
It's already a bit awkward to use, because all the paths in it will be relative to where the actual file is, so we can't have things like `"output": "./lib"` there, since it'll be relative to the root, so those fields would have to be overriden in every package anyway.  
But even this way it didn't work for me, tsc produced some weird compilation error about not finding d.ts files, I think also related to relative paths.

So for now I'll leave as it is, but maybe we'll find a way to unify them.

packages/frint/src/App.ts
markvincze: Isn't it nicer if it stays inside the function? I mean this is a local helper just to be able to do recursion, right? So it's really local to this function.

markvincze:
@fahad19 fixed the comments, or wrote answers. Besides the tests, do we have an established way to test E2E if things still work, in various browsers, etc.?

fahad19:
> Besides the tests, do we have an established way to test E2E if things still work, in various browsers, etc.?

We don't actually. There is a `repl` directory in the root though, that we occasionally use for testing in real browsers against the generated dist files.

packages/frint/src/App.ts
```diff
@@ -0,0 +1,418 @@
+import { Container, createContainer, Provider, resolveContainer } from 'frint-di';
+import find = require('lodash/find');
+import findIndex = require('lodash/findIndex');
```
fahad19: is this a TypeScript specific syntax? I am used to seeing imports like this:

```js
import findIndex from 'lodash/findIndex'; 

// OR

import * as findIndex from 'lodash/findIndex';
```

packages/frint/src/App.ts
```diff
@@ -0,0 +1,418 @@
...
+import get = require('lodash/get');
+import omit = require('lodash/omit');
+import 'rxjs/add/observable/of';
```
fahad19: `rxjs/add/*` has side-effects, and mutates the package globally.

what's happening here:

```js
// rxjs/add/observable/of.js

// pure side-effects free modules
const Observable = require('rxjs/Observable');
const of = require('rxjs/observable/of');

// mutation
Observable.of = of;
```

that's why we added an ESLint rule here to avoid these imports: https://github.com/frintjs/frint/blob/master/.eslintrc#L12

may be we can do something similar with TSLint?

packages/frint/src/App.ts
```diff
@@ -0,0 +1,418 @@
...
+    return instance;
+  }
+
+  public getAppOnceAvailable$(name, region = null, regionKey = null) {
+    const rootApp = this.getRootApp();
+
+    const w = rootApp.getAppInstance(name, region, regionKey);
+
+    if (w) {
+      return Observable.of(w);
```
fahad19: best to import the `of` module separately:

```js
import { of } from 'rxjs/observable/of';

return of(w);
```

packages/frint/src/App.ts
discosultan: > for every call, function B will be defined again and again, and then called

Huh, I'm very curious about this one :)

packages/frint/src/App.ts
markvincze: Yes, these don't work in TypeScript, I found this alternative way here: https://github.com/lodash/lodash/issues/3192

If I do
```
import omit from 'lodash/omit';
```
that gives me this error: `[ts] Module '"frint/node_modules/@types/lodash/omit"' has no default export.`  
On the other hand, it works with `rxjs`, I think because they do exports differently.  
Rxjs does (in `find.d.ts`)
```
export declare function find...
```
And `lodash` is doing this (in `omit.d.ts`)
```
import { omit } from "./index";
export = omit;
```

I'm not really familiar with what the difference is between these two styles, is it familiar to you?

packages/frint/src/App.ts
markvincze: Done

packages/frint/src/App.ts
markvincze: Yep, there is a `no-import-side-effect` rule, I added it to `tslint.json`.

packages/frint/src/App.ts
markvincze: I tried to test it locally with Node, and I had to increase the number of iterations to 100,000,000 until I noticed a measurable difference, so in this case the overhead is neglibible (especially since the local `findParents` is created only once per `getParentApps` call.  
In terms of memory overhead—according to what I found—a local function takes up approximately as much memory as an empty object (like `let foo = {};`).

So I wouldn't be worried about overhead here at all, I'd just choose whichever is more readable. If you prefer the other one, I can change it.

markvincze:
@fahad19 I guess we're getting close to be able to merge this. Can we sit together next week to do some e2e tests? :)

packages/frint/src/App.ts
fahad19: 👍

fahad19:
> I guess we're getting close to be able to merge this. Can we sit together next week to do some e2e tests? :)

Yes! Let's do it soon. may be tomorrow?

Brilliant effort on this whole initiative, @markvincze! we are lately receiving issues submitted by external parties regarding TypeScript/typings/etc. This will greatly help!

PR #413, opened 2018-02-20
frint-react: improve display name of observed components

rbardini:
## What's done

Add `name` as a fallback for `displayName` when defining the display name of an observed component.

## Why

Usually, we don't need to set `displayName` explicitly because it's inferred from the name of the function or class that defines the component. However, this inferred value may be set as a `name` class property instead, so we should consider this case as well.

This change follows the convention for [wrapping the display name of a wrapped component](https://reactjs.org/docs/higher-order-components.html#convention-wrap-the-display-name-for-easy-debugging) from React docs.

### DIFF
--- a/packages/frint-react/src/components/observe.js
+++ b/packages/frint-react/src/components/observe.js
@@ -11,9 +11,13 @@ import isObservable from '../isObservable';
 
 export default function observe(fn) {
   return (Component) => {
+    const componentName = (typeof Component.displayName !== 'undefined')
+      ? Component.displayName
+      : Component.name;
+
     class WrappedComponent extends React.Component {
-      static displayName = (typeof Component.displayName !== 'undefined')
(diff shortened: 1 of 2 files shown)


fahad19:
Thanks for the PR, @rbardini! 💖 

Let's sit together tomorrow for 10 mins, and I will sort out your GitHub/npm access, and go through the release process together. Then you can handle it yourself in future.

rbardini:
No problem, @fahad19! Let me know when you are free 🙂

fahad19:
in the meantime, could you please investigate what happened to the coverage? thanks!

rbardini:
I think the coverage decrease has to do with the test cases I've added: since I'm not actually rendering the test components, their 'render' logic is never covered.

I'll update the PR with simpler test components, so the coverage should go back to normal.

PR #416, opened 2018-04-06
Define type of Container.get() return value with generic

viacheslaff:
Following suggestion by @markvincze, I've changed Container.get() signature from `get(name: string): any;` to `get<T>(name: string): T;`.

This not only allows to leverage type safety of TypeScript, but requires to use it.

Previously Container type was optional

```ts
const console: Console = container.get('console');
console.log()
```

Now if you don't specify `<Console>` it'll default to `{}` (empty object type) and `console.log()` won't compile. This is not ideal but unfortunately I don't know the way to require generic type in TS.  Nevertheless it achieves the goal of making type required, because without using the type you pretty much cannot use the value received from the container.

```ts
const console = container.get<Console>('console');
console.log() // works
```

```ts
const console = container.get('console');
console.log() // error TS2339: Property 'log' does not exist on type '{}'
```

### DIFF
--- a/packages/frint-cli/src/providers.mock.ts
+++ b/packages/frint-cli/src/providers.mock.ts
@@ -1,5 +1,12 @@
 import * as MemoryFs from 'memory-fs';
 
+export interface FakeConsole {
+  errors: string[],
+  logs: string[],
+  log: (message: string) => void,
+  error: (message: string) => void,
+}
+
 const fs = new MemoryFs();
 
 export default [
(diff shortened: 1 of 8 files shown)


packages/frint-cli/src/providers.mock.ts
```diff
@@ -36,7 +43,7 @@ export default [
     name: 'console',
     cascade: true,
     useFactory: function useFactory() {
-      const fakeConsole = {
+      const fakeConsole: FakeConsole = {
```
fahad19: native `Console` not supported? https://github.com/Microsoft/TypeScript/blob/master/src/lib/dom.generated.d.ts#L2983-L3011

Issue #412, opened 2018-02-14
Migrate to TypeScript


fahad19:
## Packages

* [x] `frint-cli`
* [ ] `frint` (#406)

PR #421, opened 2018-05-25
Typescript fix: export function Region

andrewparadi:
I started working on a FrintJS app using the `multiple-apps` as starter code. I began by moving the repo to Typescript and encountered this persistent error in `Root.jsx` when Region is imported.

```
Module '"/Users/aparadi/Development/misk-admin/main/node_modules/frint-react/index"' has no exported member 'Region'.
```

To fix this, Region is added as an export to the index definition file and the error no longer appears. 

I'm still early in my Typescript days so if this isn't the best way to resolve the error, please comment below.


fahad19:
thanks for your valuable contribution, @andrewparadi! 

could you please also try something like this as mentioned in this comment? the more strict the better :) https://github.com/frintjs/frint/issues/410#issuecomment-363776194

we also have a placeholder issue https://github.com/frintjs/frint/issues/412 for migrating all our packages to TypeScript. if you feel like taking a stab at them, please feel free :)

andrewparadi:
Thanks for the heads up to those issues, I'll plan to make other PRs under #412 as I come across any other Typescript issues in the weeks to come in my development.

A huge thank you already for FrintJS, exactly what we've been searching for, very thankful I stumbled upon it instead of having to try and build it all from scratch.

andrewparadi:
Commit squashed and ready for merge if no other changes are requested.

fahad19:
@frintjs/core: please take a look and approve :D

fahad19:
@leandrooriente: please let me know if you would be willing to merge and release a new patch version.

PR #423, opened 2018-06-11
[#412] frint-config/Typescript

andrewparadi:
Unsure how to get past the below typing error so I added rudimentary Typescript support just with `any` for function-returned arrays of externals.

Issue
---
Consider if there was a type `External`.

```Typescript
declare type External = {
  root: string|string[]|Array<string>;
  commonjs: string;
  commonjs2: string;
  amd: string;
}
```

From my research, there doesn't seem to be a way to define a further type that is an object of strings mapping to External.

```Typescript
let lodashExternal : External = {
      root: '_',
      commonjs: 'lodash',
      commonjs2: 'lodash',
      amd: 'lodash',
   }

let rxjsExternal : External = { ... }

export const thirdParties = [{
  'lodash': lodashExternal,
  'rxjs': rxjsExternal,
    ...
]
```

In the example above, though it is clear that the `const thirdParties` consists of named External objects, how can this type be defined without explicitly defining each library?

```Typescript
export declare type Externals = [{
  'lodash'?: External,
  'rxjs'?: External,
  'react'?: External,
  'react-dom'?: External,
  'prop-types'?: External,
  'frint'?: External,
  'frint-store'?: External,
  'frint-model'?: External,
  'frint-data'?: External,
  'frint-react'?: External,
  'frint-router'?: External,
  'frint-router-react'?: External,
...
}
```


fahad19:
Thanks for the PR, @andrewparadi!

TypeScript also supports `any` for keys. for example, in this interface:

```ts
interface Foo {
  [key: string]: any;
}
```

It means, any object implementing the `Foo` interface can have any keys which is a `string`, which `any` value.

fahad19:
Refs #412

PR #427, opened 2018-07-10
frint-react: add `hydrate` function to `frint-react`

### LINKED ISSUE #335: frint-react: support hydration from server-rendered HTML
## Currently

We can only render in the browser targeting a specific node.

## Scenario for improvement

When HTML is rendered from the server, we don't really need to render everything again in the browser, since React can pick up from where the server rendering ended.

This can be achieved with `ReactDOM.hydrate`: https://reactjs.org/docs/react-dom.html#hydrate

Adding a new function in `frint-react` for `hydrate` would help achieve a nice solution for server-to-client flow.

rbardini:
## What's done

Add `hydrate` function to `frint-react`.

## Why

Support hydration from server-rendered HTML.

Closes #335.

### DIFF
--- a/packages/frint-react/README.md
+++ b/packages/frint-react/README.md
@@ -16,6 +16,7 @@
   - [Multi-instance Apps](#multi-instance-apps)
 - [API](#api)
   - [render](#render)
+  - [hydrate](#hydrate)
   - [observe](#observe)
   - [Region](#region)
   - [RegionService](#regionservice)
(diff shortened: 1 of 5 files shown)


fahad19:
Great going, @rbardini! 💖

fahad19:
please also update the README of `frint-react` 😬

rbardini:
Totally forgot about it! Updated 🙂

armand1m:
wow, nice! @rbardini great work!

Issue #418, opened 2018-04-25
Extract component handlers packages out of monorepo


fahad19:
## Currently 

We have:

* `frint-component-handlers`
* `frint-router-component-handlers`

Which are used respectively by:

* `frint-react`
* `frint-router-react`

We did it that way to enable us to create other packages integrating with various rendering libraries (like Preact, Vue, and others).

## Proposal

We haven't done any further maintenance of other integration packages, besides the one for React which exists in the monorepo.

I suggest moving the handler packages out of the monorepo, so other integration packages can still use them. 

But the React integration packages that we have in the monorepo don't need to use them any more to make lives of core maintainers easier.

PR #429, opened 2018-07-11
Added support for SSR on regions and refactored to remove handlers

jcampalo:


fahad19:
Refs #418

fahad19:
it would be useful if we also add tests in `frint-react-server`, where we `renderToString()` an App with a Child App.

fahad19:
top notch work, @jcampalo! 💖 

many thanks for fixing this :D 

please feel free to collaborate with @rbardini regarding the merge and release. he can sort out your npm access too if needed.

packages/frint-react-server/src/renderToString.spec.js
```diff
@@ -7,6 +7,7 @@ import { createApp } from 'frint';
 import { observe, streamProps } from 'frint-react';
 
 import renderToString from './renderToString';
+import Region from '../../frint-react/src/components/Region';
```
fahad19: if you do `npm run bootstrap` from root, you can import modules directly from `frint-react` instead of relative path within the monorepo.

fahad19:
Coverage dropped below 97%, that's why the build is failing: https://travis-ci.org/frintjs/frint/builds/402709904#L5487

packages/frint-react/package.json
```diff
@@ -34,11 +34,17 @@
...
     "frint": "^5.5.0",
     "frint-config": "^5.5.0",
-    "frint-test-utils": "^5.5.0"
+    "frint-test-utils": "^5.5.0",
+    "frint-react-server": "^5.5.0"
   },
   "bugs": {
     "url": "https://github.com/frintjs/frint/issues"
   },
+  "nyc": {
```
fahad19: 👍

PR #431, opened 2018-07-12
frint: Expose class methods on the app instance level

### LINKED ISSUE #426: Proposal: Expose class methods on the app instance level
## Currently

We can declare methods inside `createApp` to expose application-specific behavior:

```js
const FooApp = createApp({
  name: 'Foo',
  bar() { // exposed method
    console.log('bar')
  }
})
```

This requires accessing the `options` property of the app instance to retrieve the methods from another app:

```js
const BazComponent = observe(app =>
  streamProps()
    .set({
      foo: app.get('foo').options.bar, // retrieve `bar` method out of `foo` app options
    })
    .get$()
)(({ bar }) => /* `bar` method as prop */)

const BazApp = createApp({
  name: 'Baz',
  providers: [
    {
      name: 'component',
      useValue: BazComponent,
    },
    {
      name: 'foo',
      deps: ['rootApp'],
      useFactory({ rootApp }) {
        return rootApp.getAppInstance('Foo')
      }
    }
  ]
})
```

## Proposal

Add a new, optional `methods` (`Object`) property to the `options` argument of `createApp`, which has its properties added to the app's `this` object, thus allowing us to access the methods directly from the app instance:

```js
const FooApp = createApp({
  name: 'Foo',
  methods: { // specify what methods this app exposes
    bar() {
      console.log('bar')
    }
  }
})
```

```js
const BazComponent = observe(app =>
  streamProps()
    .set({
      foo: app.get('foo').bar, // retrieve `bar` method out of `foo` app directly
    })
    .get$()
)(({ bar }) => /* `bar` method as prop */)

const BazApp = createApp({
  name: 'Baz',
  providers: [
    {
      name: 'component',
      useValue: BazComponent,
    },
    {
      name: 'foo',
      deps: ['rootApp'],
      useFactory({ rootApp }) {
        return rootApp.getAppInstance('Foo')
      }
    }
  ]
})
```

rbardini:
## What's done

- Add optional `methods` (`Object`) property to `AppOptions`, which has its properties added to the app's `this` object.
- Add `app.getMethods` method which returns the app's `methods` object.

## Why

Allow access to public methods directly from the app instance, besides the `options` property.

Closes #426.

### DIFF
--- a/packages/frint/src/App.ts
+++ b/packages/frint/src/App.ts
@@ -93,7 +93,7 @@ export interface AppClass {
 }
 
 export class App {
-  [property: string]: any;
+  [field: string]: any;
   public container: Container;
   private options: AppOptions;
   private _appsCollection: AppRegistration[];
@@ -128,6 +128,10 @@ export class App {
       const method = this.options.methods[methodName];
 
       if (typeof method === 'function') {
+        if (this[methodName] !== undefined) {
+          throw new Error(`Cannot overwrite app's \`${methodName}\` property or method with options method.`);
+        }
+
         this[methodName] = method.bind(this);
       }
     });
--- a/packages/frint/src/App.ts
+++ b/packages/frint/src/App.ts
@@ -32,6 +32,10 @@ function makeInstanceKey(region = null, regionKey = null, multi = false) {
   return key;
 }
 
+export interface Methods {
+  [key: string]: () => any;
+}
+
 export interface ProviderNames {
   component: string;
   container: string;
@@ -75,6 +79,7 @@ export interface AppRegistration {
 
 export interface AppOptions {
   name?: string;
+  methods?: Methods;
   parentApp?: App;
   providers?: FrintProvider[];
   providerNames?: ProviderNames;
@@ -88,6 +93,7 @@ export interface AppClass {
 }
 
 export class App {
+  [property: string]: any;
   public container: Container;
   private options: AppOptions;
   private _appsCollection: AppRegistration[];
@@ -96,6 +102,7 @@ export class App {
   constructor(opts: AppOptions) {
     this.options = {
       name: null,
+      methods: {},
       parentApp: null,
       providers: [],
 
@@ -115,6 +122,15 @@ export class App {
     if (!this.options.name) {
       throw new Error('Must provide `name` in options');
     }
+
+    // expose methods as class properties
+    Object.keys(this.options.methods).forEach(methodName => {
+      const method = this.options.methods[methodName];
+
+      if (typeof method === 'function') {
+        this[methodName] = method.bind(this);
+      }
+    });
 
     // children - create Observable if root
     this._appsCollection = [];
@@ -181,6 +197,10 @@ export class App {
 
   public getName() {
     return this.getOption('name');
+  }
+
+  public getMethods() {
+    return this.getOption('methods');
   }
 
   public getProviders() {
--- a/packages/frint/src/App.ts
+++ b/packages/frint/src/App.ts
@@ -93,7 +93,7 @@ export interface AppClass {
 }
 
 export class App {
-  [field: string]: any;
+  [method: string]: any;
   public container: Container;
   private options: AppOptions;
   private _appsCollection: AppRegistration[];
--- a/packages/frint/src/App.ts
+++ b/packages/frint/src/App.ts
@@ -203,10 +203,6 @@ export class App {
     return this.getOption('name');
   }
 
-  public getMethods() {
-    return this.getOption('methods');
-  }
-
   public getProviders() {
     return this.options.providers;
   }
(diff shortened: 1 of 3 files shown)


packages/frint/src/App.ts
```diff
@@ -183,6 +199,10 @@ export class App {
     return this.getOption('name');
   }
 
+  public getMethods() {
```
fahad19: is this new method necessary?

noisae: I think so if for some reason we change the way we save and get those methods, we don't need to change the implementation. Exposing this clear api we avoid future maintainability problems. What do you think?

fahad19: I am trying to convince myself to keep this, but not yet there :D 

at the moment, it is returning the original `methods` option that is passed to the App's constructor. those functions are not bound to the `app` instance, and cannot be called directly either. so don't know how they can be useful right away.

packages/frint/src/App.ts
```diff
@@ -116,6 +123,15 @@ export class App {
       throw new Error('Must provide `name` in options');
     }
 
+    // expose methods as class properties
+    Object.keys(this.options.methods).forEach(methodName => {
+      const method = this.options.methods[methodName];
+
+      if (typeof method === 'function') {
+        this[methodName] = method.bind(this);
```
fahad19: it will be useful if we check if there is already an existing method with the same name first.

if it exists, we throw an Error.

packages/frint/src/App.ts
noisae: I got your point, i thought it was about the way we are exposing this but i agree now that maybe we don't need this method.

rbardini: Maybe it would be useful for someone who wants to retrieve a list of the public methods exposed by a given app. But yes, not having the methods bound to the app instance makes the return value less useful—and it's accessible via `app.getOption('methods')` anyway.

packages/frint/src/App.ts
rbardini: Good idea, will do 🙂

packages/frint/src/App.ts
fahad19: yep :)

rbardini: I'll remove it for now to avoid confusion. We can write a proposal later on if necessary 🙂

noisae: But i don't know, i think we are doing some magic here, seems like a bad smell to me, we now are binding those methods and changing the scope, so we are exposing app instance inside each function, is that the desired behavior?

What do you guys think?

```javascript
const FooApp = createApp({
  name: 'Foo',
  methods: { // specify what methods this app exposes
    bar() {
      this.methods = {}; // exposing app variable
      console.log('bar')
    }
  }
})
```

packages/frint/src/App.ts
```diff
@@ -116,6 +123,19 @@ export class App {
...
+    // expose methods as class properties
+    Object.keys(this.options.methods).forEach(methodName => {
+      const method = this.options.methods[methodName];
+
+      if (typeof method === 'function') {
+        if (this[methodName] !== undefined) {
+          throw new Error(`Cannot overwrite app's \`${methodName}\` property or method with options method.`);
+        }
+
+        this[methodName] = method.bind(this);
```
noisae: I don't know, i think we are doing some magic here, seems like a bad smell to me, we now are binding those methods and changing the scope, so we are exposing app instance inside each function, is that the desired behaviour?

What do you guys think?

const FooApp = createApp({
  name: 'Foo',
  methods: { // specify what methods this app exposes
    bar() {
      this.methods = {}; // exposing app variable
      console.log('bar')
    }
  }
})

rbardini: That's an interesting point, and I can see how this may give a lot of freedom to break things. On the other hand, if the framework doesn't automatically bind the methods, the developer could still do and have the same privileges, so I'm not sure if this is a big problem 🤔

noisae: But there is another problems here too, if we need to create new methods on this class, and someone is using that name, when we update frintjs, this application is going to break too. We are also breaking the Open Close principle here i think.

fahad19:
Great stuff, @rbardini! 💖

packages/frint/src/App.ts
noisae: We are really open to modification on this case. 
But the thing is, imagine that i create a method called getMethods on my application, and then we do the same on App. Now when i update my frint package my application is going to break.

I think this is not a good thing :/

rbardini: One possible drawback I foresee is that, once this is merged, any new properties or methods added to the `App` class would require a major release, since they can potentially clash with injected methods. Should we worry about this, or is the `App` API stable enough for this to not be an issue?

fahad19: we haven't added any new methods for months, if not years to the App class.

PR #436, opened 2018-07-16
frint-react:  Forward `Region` props to children components

rbardini:
## What's done

Forward all Region props but `children` and `className` to the App's component by default.

## Why

Make it easier to access Region props from an App's component, without having to setup observers and forward the props yourself.

### DIFF
--- a/packages/frint-react/README.md
+++ b/packages/frint-react/README.md
@@ -301,7 +301,9 @@ We have just made our simple Component reactive, by wrapping it with `observe`. 
 
 ### Observing Region's data
 
-In previous example, we showed you how to access Region's data via `RegionService`. Now let's see how we can pass it to your App's component too:
+In previous example, we showed you how to access Region's data via `RegionService`. Now let's see how we can pass it to your App's component too.
+
+All Region props but `children` and `className` are forwarded to the App's component by default. However, if you prefer a reactive approach:
 
 ```js
 import { map } from 'rxjs/operators/map';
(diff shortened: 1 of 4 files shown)


packages/frint-react/README.md
```diff
@@ -301,7 +301,9 @@ We have just made our simple Component reactive, by wrapping it with `observe`.
 
 ### Observing Region's data
 
-In previous example, we showed you how to access Region's data via `RegionService`. Now let's see how we can pass it to your App's component too:
+In previous example, we showed you how to access Region's data via `RegionService`. Now let's see how we can pass it to your App's component too.
+
+All Region props but `children` and `className` are forwarded to the App's component by default. However, if you prefer a reactive approach:
```
fahad19: 👍

fahad19:
Great work, @rbardini! 💖

packages/frint-react/src/components/Region.js
```diff
@@ -148,18 +145,14 @@ export default class Region extends React.Component {
 
   componentWillReceiveProps(nextProps) {
     const {
-      name = this.props.name,
-      uniqueKey = this.props.uniqueKey,
-      data = this.props.data,
```
rbardini: @fahad19 do you know if we need these defaults based on the current props? I'm assuming we don't, but maybe there is a reason for it that I'm not aware of 🤔

fahad19: I actually don't see any value in extracted them separately any more.

rbardini: Sure, they will still be present in `...props` but they won't have their default values based on `this.props` anymore. I'm wondering if this change could break something.

fahad19:
@rbardini: let's merge it? :)

PR #438, opened 2018-09-05
Fixes #437 - Add generic type for Frint Methods interface

### LINKED ISSUE #437: Method interface should accept arguments
The current interface doesn't allow any argument. https://github.com/frintjs/frint/blob/master/packages/frint/src/App.ts#L36

```
export interface Methods {
    [key: string]: () => any;
}
```

It would be more useful if this interface accepts any value as an argument.

```
export interface Methods {
    [key: string]: (...args: any[]) => any;
}
```

leandrooriente:
Extend Method Interface to receive arguments in methods.

Ex: 
```typescript
const exampleApp = createApp({
  name: 'Example',
  methods: {
    sum: (a: number, b: number): number =>  a + b
  }
});
```

### DIFF
--- a/packages/frint/src/App.ts
+++ b/packages/frint/src/App.ts
@@ -33,7 +33,7 @@ function makeInstanceKey(region = null, regionKey = null, multi = false) {
 }
 
 export interface Methods {
-  [key: string]: () => any;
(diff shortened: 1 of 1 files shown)


fahad19:
💖

